PR #9, opened 2018-11-09
Show "Move All -> New Album" at all times. Actually move the photos

d7415:
when a new album is added this way.

Fixes LycheeOrg/Lychee#38

Rewrite of LycheeOrg/Lychee#84

### DIFF
--- a/scripts/album.js
+++ b/scripts/album.js
@@ -153,7 +153,7 @@ album.add = function() {
 
 		let title = data.title;
 
-		const isNumber = (n) => (!isNaN(parseFloat(n)) && isFinite(n));
+		const isNumber = (n) => (!isNaN(parseInt(n, 10)) && isFinite(n));
 
 		basicModal.close();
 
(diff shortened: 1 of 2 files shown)


scripts/album.js
```diff
@@ -190,6 +190,50 @@ album.add = function() {
 
 };
 
+album.addandmove = function (photoIDs) {
+
+	const action = function(data) {
+
+		let title = data.title;
+
+		const isNumber = (n) => (!isNaN(parseFloat(n)) && isFinite(n));
```
ildyria: I think it should be parseInt() instead of parseFloat.

d7415: Possibly. That's taken directly from album.add (Line 156). Would you like me to change both?

ildyria: please do with `parseInt(x, 10)` (specify the base used in order to make sure that the use over different platform is the same). :)

d7415: Done :)

ildyria:
Thank you. :)

PR #21, opened 2018-12-05
Album-level licenses

bennettscience:
UI updated to display licenses for albums.
- New sidebar section for the visible album showing the current license
- `setLicense()` method added for albums from the sidebar
- Sidebar updates with new license stored in the table on load.

### DIFF
--- a/scripts/album.js
+++ b/scripts/album.js
@@ -354,10 +354,8 @@ album.setDescription = function(albumID) {
 
 album.setLicense = function(albumID) {
 
-	let oldLicense = album.json.license;
-
 	const callback = function() {
-		$('select#license').val(album.json.license === '' ? lychee.locale['ALBUM_LICENSE_NONE'] : album.json.license);
+		$('select#license').val(album.json.license === '' ? 'none' : album.json.license);
 		return false;
 	};
 
@@ -366,10 +364,6 @@ album.setLicense = function(albumID) {
 		let license = data.license;
 
 		basicModal.close();
-
-		// if (visible.album()) {
-		//
-		// }
 
 		let params = {
 			albumID,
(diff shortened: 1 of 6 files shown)


scripts/lychee_locale.js
```diff
@@ -130,6 +130,12 @@ lychee.locale = {
 		'MOVE_ALBUMS'               : "Move Albums",
 		'NOT_MOVE_ALBUMS'           : "Don't Move",
 		'ROOT'                      : "Root",
+		'ALBUM_REUSE'				: "Reuse",
+		'ALBUM_LICENSE'				: 'License',
+		'ALBUM_SET_LICENSE'			: 'Set License',
+		'ALBUM_LICENSE_HELP'		: 'Need help choosing?',
+		'ALBUM_LICENSE_NONE'		: 'None',
+		'ALBUM_RESERVED'			: 'All Rights Reserved',
```
ildyria: Don't forget to add those in the `English.php` ;)

scripts/album.js
```diff
@@ -352,6 +352,77 @@ album.setDescription = function(albumID) {
...
+	};
+
+	const action = function(data) {
+
+		let license = data.license;
+
+		basicModal.close();
+
+		if (visible.album()) {
+			album.json.license = license;
```
ildyria: this update should be done in after:
```js
if (data!==true) lychee.error(null, params, data)
```
something in the line of:
```js
if (data!==true) {
	lychee.error(null, params, data)
}
else
{
		if (visible.album()) {
			album.json.license = license;
			view.album.license()
		}
}
```

In such way you update the json only if you are sure that it has been modified in the server side.

Issue #29, opened 2019-02-15
Thumbnails displayed in albums overview are low-density


kamil4:
### Detailed description of the problem
I have a HiDPI ("retina") display. The square thumbnails displayed on each album-specific page are high-density (400x400px resolution), as expected. However, the album thumbnails displayed in the albums overview (the Lychee home page) are low-density (200x200px resolution).

### Steps to reproduce the issue
Use a HiDPI ("retina") screen. I suppose a 200% zoom on a regular screen will also work.
Compare the quality of the thumbnails in the albums overview to the ones displayed inside an album.

### Output of the diagnostics (Settings => Diagnostics)
Lychee Version (json):  3.2.12
DB Version:             update_030212
System:                 Linux
PHP Version:            7.2
MySQL Version:          50724
Imagick:                -
Imagick Active:         0
Imagick Version:        -
GD Version:             bundled (2.1.0 compatible)
Plugins:                

### Browser and system
Firefox 64.0.2, Linux

PR #30, opened 2019-02-15
Use HiDPI thumbnails in albums overview

kamil4:
Also fix the incorrect pixel density.

There were several problems here.
The major one was that even though lychee.retinize() was being correctly invoked in build.album(), the result was not being passed to build.getThumbnailHtml().
A minor one was that lychee.retinize() was only invoked for the front picture and not for the two rotated ones behind it. Being a perfectionist I fixed that as well.
Finally, the "srcset" attribute in build.getThumbnailHtml() specified wrong pixel density. Since the "@2x" thumbnails are 400x400 vs 200x200 for the regular ones, that value should obviously be "2x" but for some reason "1.5x" was being used.


kamil4:
Fixes #29

PR #31, opened 2019-02-19
Unjustified layout

kamil4:
Here's my initial attempt at implementing unjustified layout as described in https://github.com/LycheeOrg/Lychee/issues/199.

In the end I didn't need to use any third-party packages. I just needed to set the dimensions of each `<div photo>` (I originally thought that they could inherit them from the `` tags inside but I guess that's not how it works). With the correct dimensions, the browser takes care of the rest.

I didn't want to hard-code the height of individual images so instead it's set in the CSS so a user should be able to override it. I set it to 240px, which looks good to me (Google Photos has it at 220px by default, I think). Another way to do it would be to tie it to the height of small images, I guess...

I tried to keep the original code as-is as much as possible. In particular, I kept `lychee.justified` unchanged, adding a separate `lychee.unjustified` to control the new layout. If both are `false`, squares are used. If both are `true`, then I don't know what happens, because it's never supposed to be the case :-).

I turned the "Use justified layout" switch into a 3-way selector in the Settings UI. I extended the `justified_layout` field in the database from 0/1 to 0/1/2, 2 being the new unjustified layout (that's in a separate PR against Lychee-Laravel). By the look of it, that's not how it's done for other fields (a text identifier is stored) but I didn't want to deal with modifying the format of settings in the database. Let me know whether that's OK or if I should bite the bullet and keep it consistent with the rest.

### DIFF
--- a/scripts/lychee_locale.js
+++ b/scripts/lychee_locale.js
@@ -278,10 +278,15 @@ lychee.locale = {
 		'CSS_TEXT'					: 'Personalize your CSS:',
 		'CSS_TITLE'				    : 'Change CSS',
 
-		'LAYOUT_TEXT'				: 'Use justified layout:',
+		'LAYOUT_TYPE'				: 'Layout of photos:',
+		'LAYOUT_SQUARES'			: 'Square thumbnails',
+		'LAYOUT_JUSTIFIED'			: 'With aspect, justified',
+		'LAYOUT_UNJUSTIFIED'			: 'With aspect, unjustified',
+		'SET_LAYOUT'				: 'Change layout',
+
 		'IMAGE_OVERLAY_TEXT'		: 'Display image overlay by default:',
 
-		'LAYOUT_TYPE'				: 'Data to use in image overlay:',
+		'OVERLAY_TYPE'				: 'Data to use in image overlay:',
 		'OVERLAY_EXIF'				: 'Photo EXIF data',
 		'OVERLAY_DESCRIPTION'		: 'Photo description',
 		'OVERLAY_DATE'				: 'Photo date taken',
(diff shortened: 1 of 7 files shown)


scripts/lychee_locale.js
```diff
@@ -278,10 +278,15 @@ lychee.locale = {
...
+		'LAYOUT_TYPE'				: 'Layout of photos:',
+		'LAYOUT_SQUARES'			: 'Square thumbnails',
+		'LAYOUT_JUSTIFIED'			: 'With aspect, justified',
+		'LAYOUT_UNJUSTIFIED'			: 'With aspect, unjustified',
+		'SET_LAYOUT'				: 'Change layout',
+
 		'IMAGE_OVERLAY_TEXT'		: 'Display image overlay by default:',
 
-		'LAYOUT_TYPE'				: 'Data to use in image overlay:',
+		'OVERLAY_TYPE'				: 'Data to use in image overlay:',
```
ildyria: Nicely spotted. :+1:

scripts/settings.js
```diff
@@ -340,27 +340,18 @@ settings.setDefaultLicense = function(params) {
 		if (data===true) {
 			lychee.default_license = params.license;
 			loadingBar.show('success', lychee.locale['SETTINGS_SUCCESS_LICENSE']);
-		} else lychee.error(ull, params, data)
+		} else lychee.error(null, params, data)
```
ildyria: Nicely spotted. :+1:

ildyria:
Double checking.

@kamil4 is this your desired result?
2019-02-19_1920x1080_11 30 59

kamil4:
Yep, that looks exactly as intended.

Let me backport the server changes to Lychee v3 though. I'll submit a separate PR for that.

PR #32, opened 2019-02-19
Clean up justified/unjustified layouts. Fix some more whitespace.

d7415:


d7415:
See also LycheeOrg/Lychee#212

d7415:
(Also, please either straight merge or wait for me as I've used the refs in the Lychee PR)

d7415:
(Laravel awaiting me figuring it out :P)

d7415:
> value needs to be "2"

Good catch. I told @ildyria I'd create bugs!

scripts/view.js
```diff
@@ -205,10 +205,10 @@ view.album = {
 			}
 
 			if (photosData !== '') {
-				if (lychee.justified) {
+				if (lychee.layout == '1') {
 					photosData = '<div class="justified-layout">' + photosData + '</div>';
 				}
-				else if (lychee.unjustified) {
+				else if (lychee.layout == '2') {
```
ildyria: are you sure it should not be `===` instead of just `==` ?

scripts/view.js
```diff
@@ -231,7 +231,7 @@ view.album = {
 			// Add photos to view
 			lychee.content.html(html);
 			view.album.content.justify();
-			if (lychee.unjustified) {
+			if (lychee.layout == '2') {
```
ildyria: are you sure it should not be `===` instead of just `==` ?

scripts/view.js
```diff
@@ -288,7 +288,7 @@ view.album = {
 		},
 
 		justify: function () {
-			if (!lychee.justified) return;
+			if (lychee.layout != '1') return;
```
ildyria: are you sure it should not be `!==` instead of just `!=` ?

d7415:
I think you're right. I'll change it (and the dist/submodules) when I have some time later.

Thanks!

PR #34, opened 2019-02-21
Resize image height with unjustified to fit

kamil4:
I was hoping that the unjustified layout could be simple, static HTML with fixed image sizes, but alas... I forgot about extremely narrow browser windows. In that case the images do need to be rescaled to make them fit within the available space, and it has to be done dynamically in case the window width is changed.

So I moved the unjustified layout code from init() to justify() and expanded it to handle rescaling if needed, based on current container width.


ildyria:
@kamil4 Sorry, I did a lot of moves between files in the last PR. Can you solve the conflicts ? (it should just be moving files around)

kamil4:
I rebased my changes and pushed them to this PR branch, hopefully correctly (sorry, I'm still learning how to use git in collaborative settings). Let me know if this is fine; otherwise I guess I'll create a new branch with a single, clean commit and send a new PR.

kamil4:
Oh, I see you merged it while I was typing this.  Never mind then :-).

ildyria:
:sunglasses:

PR #37, opened 2019-02-27
Frontent part of full HiDPI support

kamil4:
This merge request addresses https://github.com/LycheeOrg/Lychee/issues/198.  See https://github.com/LycheeOrg/Lychee-Laravel/pull/92 for the server part and the overview/motivation of why things are done the way they are.

As implement, support for HiDPI requires extensions to the server API. However, I kept the extensions optional so the new frontend can be used with Lychee v3 without any loss in existing functionality (I only implemented the server changes for the v4).

I thought I would discuss some parts of the patch:

I got rid of `build.getThumbnailHtml()` because extending it to handle HiDPI small/medium would require something like 8 new arguments.  So I replaced it with a much simpler `build.getAlbumThumb()` that's used just for the square thumbnails in albums view, and the rest of the functionality is inlined in `build.photo()` since that was the only user of it.

I added a heuristics to the prefetcher to preload the same variant of the image as the one currently displayed.  The browser chooses which variant to display based on the window size and the variant sizes passed in `srcset` and exports the chosen one via the `currentSrc` DOM property.

When the browser window is resized, the `sizes` tags of the displayed images need to be recalculated to ensure that the browser uses the optimal variant for rendering.  For album view I just extended `view.album.content.justify()`, but for image view I needed to add a new `view.photo.onresize()` method since there wasn't one before.

### DIFF
--- a/scripts/main/build.js
+++ b/scripts/main/build.js
@@ -4,43 +4,43 @@
 
 build = {};
 
-build.iconic = function(icon, classes = '') {
+build.iconic = function (icon, classes = '') {
 
 	let html = '';
 
-	html += lychee.html`<svg class='iconic ${ classes }'><use xlink:href='#${ icon }' /></svg>`;
+	html += lychee.html`<svg class='iconic ${classes}'><use xlink:href='#${icon}' /></svg>`;
 
 	return html
 
 };
 
-build.divider = function(title) {
+build.divider = function (title) {
 
 	let html = '';
 
-	html += lychee.html`<div class='divider'><h1>${ title }</h1></div>`;
+	html += lychee.html`<div class='divider'><h1>${title}</h1></div>`;
 
 	return html
 
 };
 
-build.editIcon = function(id) {
+build.editIcon = function (id) {
 
 	let html = '';
 
-	html += lychee.html`<div id='${ id }' class='edit'>${ build.iconic('pencil') }</div>`;
+	html += lychee.html`<div id='${id}' class='edit'>${build.iconic('pencil')}</div>`;
 
 	return html
 
 };
 
-build.multiselect = function(top, left) {
+build.multiselect = function (top, left) {
 
-	return lychee.html`<div id='multiselect' style='top: ${ top }px; left: ${ left }px;'></div>`
+	return lychee.html`<div id='multiselect' style='top: ${top}px; left: ${left}px;'></div>`
 
 };
 
-build.getAlbumThumb = function(data, i) {
+build.getAlbumThumb = function (data, i) {
 	let isVideo = data.types[i] && data.types[i].indexOf('video') > -1;
 	let thumb = data.thumbs[i];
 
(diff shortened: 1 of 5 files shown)


ildyria:
@kamil4  Can you check if the changes I made breaks anything on your side ? I cannot test the retina support.

d7415:
> @kamil4 Can you check if the changes I made breaks anything on your side ? I cannot test the retina support.

I think a bunch of those changes are things I was thinking about while. It's still a *lot* of lines, and passing the whole of `data` seems messy but I don't see a better way.

scripts/main/build.js
```diff
@@ -128,29 +116,83 @@ build.album = function(data, disabled = false) {
...
+			thumb2x = data.thumb2x
+		} else { // Lychee v3
+			let {path: thumb2x} = lychee.retinize(data.thumbUrl)
+		}
+
+		if (thumb2x !== '') {
+			thumb2x = `srcset='${thumb2x} 2x'`
+		}
+
+		thumbnail = `<span class="thumbimg${isVideo ? ' video' : ''}">`;
```
d7415: There are a few places here where it checks for videos to add the video class, but doesn't add the play icon. Am I missing something, or should this be all-or-nothing?

ildyria:
The nice thing is because it is backward compatible, you don't need to apply the PR on Lychee-laravel to test the Lychee-v3 behaviour. :+1:

scripts/main/build.js
ildyria: it should add the play icon if the video thumbnail is not available.

d7415: Ah, of course!

kamil4:
@ildyria, I share your taste in whitespace :wink:

One of your changes broke the image view; my latest commit should fix it.

BTW, I had no idea that you can push commits to my fork. The things you learn... :astonished:

Also, is there some JS validator you guys use/recommend? I consider myself first and foremost a C programmer so I'm missing that girdle of a picky compiler that would warn me about uninitialized variables and such :wink:

ildyria:
@kamil4 I'm using https://www.jetbrains.com/phpstorm/
It has a bunch of auto indent thing and also statically checks for non used variables, and warns you against `==`.

d7415:
> BTW, I had no idea that you can push commits to my fork. The things you learn...

There's a checkbox (default on) when you create a PR to allow the repo owner to do this :)

ildyria:
> There's a checkbox (default on) when you create a PR to allow the repo owner to do this :)

But it works the best when you actually create a new branch, because otherwise you have a "master" name collision...

kamil4:
I'm a little worried that this is not merged yet https://github.com/LycheeOrg/Lychee-Laravel/pull/92 is. I never tested "new" server with "old" frontend... Then again, if that combo didn't work, I'm sure we would've heard by now. :wink:

Anyway, is something blocking this PR? Without it the new high-res images won't get used for anything...

PR #41, opened 2019-03-04
Support for small thumbs and metadata for videos

kamil4:
Addresses https://github.com/LycheeOrg/Lychee-Laravel/issues/93. The bulk of the work is on the server side in https://github.com/LycheeOrg/Lychee-Laravel/pull/110.
Here we just add support for small (and medium, currently unused) video thumbnails and including video-specific metadata in the sidebar.


scripts/main/sidebar.js
```diff
@@ -328,12 +355,19 @@ sidebar.createStructure.album = function(data) {
...
+		}
+	});
 	structure.album = {
 		title : lychee.locale['ALBUM_ALBUM'],
 		type  : sidebar.types.DEFAULT,
 		rows  : [
 			{ title: lychee.locale['ALBUM_CREATED'], kind: 'created',       value: data.sysdate },
-			{ title: lychee.locale['ALBUM_IMAGES'],  kind: 'images',        value: data.photos.length }
+			{ title: lychee.locale['ALBUM_IMAGES'],  kind: 'images',        value: data.photos.length - videoCount },
+			{ title: lychee.locale['ALBUM_VIDEOS'],  kind: 'videos',        value: videoCount }
```
ildyria: only show line 370 if videoCount > 0.

scripts/main/sidebar.js
```diff
@@ -117,6 +117,15 @@ sidebar.changeAttr = function(attr, value = '-', dangerouslySetInnerHTML = false
 
 };
 
+sidebar.secondsToHMS = function(d) {
+	d = Number(d);
+	var h = Math.floor(d / 3600);
+	var m = Math.floor(d % 3600 / 60);
+	var s = Math.floor(d % 3600 % 60);
```
ildyria:
```php
// we have
d = h * 3600 + m * 60 + s
d % 3600 = m * 60 + s
d % 3600 % 60 = s
// but also
d = (h * 60 + m ) * 60 + s
d % 60 = s
// thus
d % 3600 % 60 = d % 60
```
:)

scripts/main/sidebar.js
```diff
@@ -117,6 +117,15 @@ sidebar.changeAttr = function(attr, value = '-', dangerouslySetInnerHTML = false
...
+
+	return ((h > 0) ? h + 'h' : '') + ((m > 0) ? m + 'm' : '') + ((s > 0) ? s + 's' : '');
```
ildyria: I would use
`.toString()` to make sure of the conversion of `h`, `m` and `s`

scripts/main/sidebar.js
kamil4: Will do...

scripts/main/sidebar.js
kamil4: That's what happens when I copy code from Stackoverflow without thinking... :wink: Good catch!

scripts/main/sidebar.js
kamil4: Done. While I was testing it, I noticed that the image count was misbehaving for empty albums so I fixed that as well.

scripts/main/sidebar.js
kamil4: Done

scripts/main/sidebar.js
kamil4: Done. I also fixed a case of sub-0s video where it should still print "0s" rather than nothing.

PR #45, opened 2019-03-06
Add HiDPI support to Frame mode

kamil4:
Front end component of https://github.com/LycheeOrg/Lychee-Laravel/pull/118.

Server now returns the whole Photo object. I extract the data for the `srcset` attribute from it. I also added a resize callback to update the `sizes` attribute (the Photo object is stored as `frame.photo` so that the callback can access it; it only really needs it for the aspect ratio).

### DIFF
--- a/scripts/frame/main.js
+++ b/scripts/frame/main.js
@@ -102,14 +102,30 @@ frame.next = function () {
 frame.refreshPicture = function () {
 	api.post('Photo::getRandom', {}, function (data) {
 		if (!data.url) console.log('URL not found');
-		if (!data.thumb) console.log('Thumb not found');
+		if (!data.thumbUrl) console.log('Thumb not found');
 
-		$('#background').attr("src", data.thumb).on("load", function () {
+		$('#background').attr("src", data.thumbUrl).on("load", function () {
 			frame.start_blur();
 		});
 
-		$('#picture').attr("src", data.url).css('display', 'inline');
+		srcset = '';
+		this.frame.photo = null;
+		if (data.medium != '') {
+			src = data.medium;
+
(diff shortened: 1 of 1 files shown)


scripts/frame/main.js
```diff
@@ -125,6 +141,21 @@ frame.set = function (data) {
 	frame.refreshPicture();
 };
 
+frame.resize = function() {
```
ildyria: why do you need this ? This should be handled by the CSS. Is it not ?

kamil4: CSS handles the layout but is _not_ used for determining which image variant from `srcset` is to be used. For that we need to provide the browser with a (relatively) accurate rendered image width in the `sizes` attribute; and the image width can obviously change when the viewport is resized, so we recalculate. We have the same sort of code for the regular photo view, BTW (in `view.photo.onresize()`).

At the risk of talking about things you may be perfectly familiar with, allow me to expand as it was something that surprised _me_ about `srcset` when I was reading the specs and experimenting with it in the browsers. Basically, we provide the URLs and widths of the image variants in the `srcset` attribute, but that's not enough for the browser's rendering engine. The browser also needs the expected rendered width to be provided, via `sizes`. Apparently that's because images are loaded before the browser determines the page layout, so at the time they are loaded the browser doesn't know what the width of the rendered image is going to be. So by default it assumes the worst-case scenario of `100vw`, potentially resulting in the loading of a larger image variant than is actually needed. E.g., in case of the frame mode displayed full-screen on a 16:9 display, images in portrait will use less than half of the viewport width. The browser doesn't know it without loading the image first, but we do because we have the width/height information in the database so we have a good idea of how wide it will be rendered. That's the purpose of the math that the code does here.

PR #51, opened 2019-03-14
Subalbums passwords

kamil4:
Front-end-side fixes for https://github.com/LycheeOrg/Lychee-Laravel/issues/139

The change to `album.js` makes `albums.getByID()` search recursively through subalbums. It's kind-of amazing how everything else worked fine without it :smiley: 

The change to `password.js` removes a shortcut that appears to be bogus? I'm not sure why the code used to check the password-protection status of the parent album when determining whether to ask for the password of a subalbum. As a result, if the parent album was not password-protected, it would never ask for the password for the subalbum, so a password-protected subalbum could not be opened.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -103,7 +103,7 @@ album.load = function(albumID, refresh = false) {
 		password: ''
 	};
 
-	processData = function(data) {
+	const processData = function (data) {
 
 		if (data === 'Warning: Wrong password!') {
 			// User hit Cancel at the password prompt
@@ -130,7 +130,7 @@ album.load = function(albumID, refresh = false) {
 
 		// Skip delay when refresh is true
 		// Skip delay when opening a blank Lychee
-		if (refresh === true)                                          waitTime = 0;
+		if (refresh === true) waitTime = 0;
 		if (!visible.albums() && !visible.photo() && !visible.album()) waitTime = 0;
 
 		setTimeout(() => {
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -98,64 +98,68 @@ album.getParent = function() {
 
 album.load = function(albumID, refresh = false) {
 
-	password.get(albumID, function() {
+	let params = {
+		albumID,
+		password: ''
+	};
 
-		if (refresh===false) lychee.animate('.content', 'contentZoomOut');
+	processData = function(data) {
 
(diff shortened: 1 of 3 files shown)


kamil4:
OK, here's a second and much more thorough attempt at this.

Thanks to @ildyria I finally understood what the code in `password.get` was attempting to do. My conclusion is that while that code was probably correct in v3, with v4 it is hopelessly broken. With subalbums and multiple users of varying privileges, the front end shouldn't try to second-guess the server; it simply doesn't have enough information.

In the end I refactored `album.load` so that instead of relying on the code in `password.get` to guess whether the front end should ask for a password or not, it right away sends the `Album::get` request to the server. If the server replies with `Warning: Wrong password!`, the front end asks for the password using `password.getDialog` and re-issues the `Album::get` request, this time with the password. This should be much more robust: it correctly handles all the nested cases I could think of and doesn't need my `passwordProvided` kludge from f24b2ee. And it seems to work fine with v3 as well.  What a deal... :smiley:

I also fixed a couple of instances of `lychee.goto()` forcing a switch to albums view if an album could not be opened. Again, this was ok in v3, but in v4, with subalbums, it should only do that if no album is being displayed. I'm sure there are more instances of that in the code and I will look for them in the future...

scripts/main/password.js
```diff
@@ -61,7 +36,7 @@ password.getDialog = function(albumID, callback) {
 	const cancel = () => {
 
 		basicModal.close();
-		if (!visible.albums()) lychee.goto()
+		if (!visible.albums() && !visible.album()) lychee.goto()
```
ildyria: nicely spotted!

scripts/main/album.js
```diff
@@ -103,7 +103,7 @@ album.load = function(albumID, refresh = false) {
 		password: ''
 	};
 
-	processData = function(data) {
+	const processData = function (data) {
```
kamil4: Out of curiosity, what does that do?

One of these days I should really go to the trouble of learning JavaScript :wink:...

ildyria: processData is implicitly declaring a variable. Thus you should use `var`. However, for a better semantic reason instead of `var` you have `let` and `const`. 

More here:
https://stackoverflow.com/questions/40070631/v8-javascript-performance-implications-of-const-let-and-var
https://ponyfoo.com/articles/var-let-const

PR #52, opened 2019-03-17
Add fullscreen support to album and photo views

kamil4:
Finally takes care of my pet project https://github.com/LycheeOrg/Lychee-Laravel/issues/94

There are also corresponding changes to the server(s). Since they are trivial, I implemented them both for v3 and v4.

I added two new buttons to album and photo views (Enter and Exit Fullscreen), only one of which is shown at any one time. I figured it was easier to do it that way than to change the button at run time.

I was wondering whether to also add the buttons to the public (and albums) views, but I thought it might be excessive, especially in the public view case where there aren't any buttons now. Let me know if you think I should add them though.

I also added the 'F' keyboard binding, as requested by @ildyria. While I was there, I fixed the 'o' keyboard binding, which was misbehaving on Firefox.

Overall it seems to work fine, even on mobile browsers (though it would be good if somebody could check on MacOS and iOS; I tested with Firefox and Chrome on Linux and Android)  but one regret I have is that it's independent/unaware of what the user did with respect to fullscreen via F11. As far as I was able to find out, it is impossible to determine that state from JS, apparently deliberately so for "security reasons". :confused:

### DIFF
--- a/scripts/main/header.js
+++ b/scripts/main/header.js
@@ -59,6 +59,8 @@ header.bind = function() {
 		}
 	});
 	header.dom('#button_back')        .on(eventName, function() { lychee.goto(album.getID()) });
+	header.dom('#button_fs_album_enter,#button_fs_enter').on(eventName, lychee.fullscreenEnter);
+	header.dom('#button_fs_album_exit,#button_fs_exit').on(eventName, lychee.fullscreenExit).hide();
 
 	header.dom('.header__search').on('keyup click', function() { search.find($(this).val()) });
 	header.dom('.header__clear').on(eventName, function() {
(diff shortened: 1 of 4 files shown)


d7415:
A much neater solution! Well done :+1:

kamil4:
Thanks :smiley:. That was the only way I could make it work, BTW. The fullscreen-exiting Esc press isn't reported to JS as a keyboard event at all...

d7415:
> The fullscreen-exiting Esc press isn't reported to JS as a keyboard event at all...

Ooh, interesting. I guess to stop websites from blocking it.

PR #53, opened 2019-03-18
Hide passwords in users view

kamil4:
Also add hover explanations for the icons/checkboxes.

Finally, tweak the layout so that columns align nicely.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -478,7 +478,7 @@ album.setPublic = function(albumID, modal, e) {
 							  <span class='label'>${ lychee.locale['ALBUM_PASSWORD_PROT'] }</span>
 						  </label>
 						  <p>${ lychee.locale['ALBUM_PASSWORD_PROT_EXPL'] }</p>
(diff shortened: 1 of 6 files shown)


ildyria:
I disagree with putting the password field as a password as we don't ask for confirmation (to prevent mistake when typing).

kamil4:
Interesting point. But the UI is inconsistent about it then. Passwords used for album protection or, especially, the admin password (asked for on the first access to a fresh installation) are hidden.

ildyria:
Then I would put the password for albums on "visible" and I would put a confirmation step for the password at a fresh installation. :)
Note that we don't ask for confirmation either for a new password when a user or the admin wants to change their password (should be the case).

In summary I would rather go for:
- Fresh installation and password change in settings: hidden with confirmation step (just another box bellow).
- User creation/modification, album password setting: visible.

Those should not need changes on the server side by the way. :)

kamil4:
OK, my latest commit should address your feedback.

And it looks like server-side changes will be needed -- due to an added string in locales.

PR #54, opened 2019-03-18
Cache album passwords in memory

kamil4:
Viewing a password-protected album can be annoying because every time you leave the album and want to re-enter it, you are asked for the password again. This is especially bad if that album has subalbums; the password needs to be reentered when going back from each subalbum to the parent album.

This can be easily prevented by caching the passwords in memory, inside the album object.

### DIFF
--- a/scripts/main/password.js
+++ b/scripts/main/password.js
@@ -12,7 +12,7 @@ password.get = function(albumID, callback) {
 
 	if (lychee.publicMode===false)                                  callback();
 	else if (album.json && album.json.password==='0')               callback();
-	else if (albums.json && albums.getByID(albumID).password==='0' || albums.getByID(albumID).passwordProvided) callback();
+	else if (albums.json && (albums.getByID(albumID).password==='0' || albums.getByID(albumID).passwordProvided)) callback();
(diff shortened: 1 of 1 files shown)


ildyria:
Hummm I think I already solved that in the Lychee-laravel version. we write the information in the session variable. If this is only a problem on the v3 I would let it as is.

https://github.com/LycheeOrg/Lychee-Laravel/blob/master/app/Http/Middleware/AlbumPWCheck.php
https://github.com/LycheeOrg/Lychee-Laravel/blob/master/app/Http/Controllers/SessionController.php#L196

kamil4:
No, it's on Lychee-Laravel. But it could be just a front end issue then. The front end decides on its own whether to ask for a password, and it does so every time a password-protected album is (re-)loaded. I guess it means we would still need to record a flag that an album has already been opened, but recording a flag would be neater than caching a password...

What should the front end pass to the server then for a reload? Just an empty string for a password?

kamil4:
OK, here's my second attempt. Indeed, the server-side in Laravel works. Which makes the front end changes much cleaner :smiley:.

PR #56, opened 2019-03-22
Hide More button in photo view if menu is empty

kamil4:
If `full_photo` is disabled in the settings and the album is not downloadable, there's no point in displaying the More button in photo view in the public mode, since the menu will be empty.


kamil4:
Sure. I basically inverted the relevant conditions from `contextMenu.photoMore` so in principle all relevant cases should be covered...

kamil4:
@ildyria can you weigh in on this PR? It's been sitting in the queue for three weeks now. Thanks! :smile:

ildyria:
oups 😄

PR #61, opened 2019-04-02
now lazy loading images in when the come in to view

### LINKED ISSUE #60: Lazy load thumbnails
When you have hundreds of photos in an album, the frontend makes a request for all of the thumbnails at once causing reduced performance. I propose that a lazy load technique is implemented so that thumbnails don't actually get requested until they are scrolled into view. This should also make the thumbnails load in faster since they're only requested when in view.

I'm more than happy to take on this issue, but wanted to make sure it was something that would be accepted first.

alex-phillips:
Fixes #60


scripts/main/build.js
```diff
@@ -139,29 +139,29 @@ build.photo = function (data) {
 		}
 
 		if (thumb2x !== '') {
-			thumb2x = `srcset='${thumb2x} 2x'`
+			thumb2x = `class="lazyload" src="dist/images/placeholder.png" data-srcset='${thumb2x} 2x'`
```
ildyria: I don't think you need the `class="lazyload"` here because it is added in line 146. :)

scripts/main/build.js
```diff
@@ -139,29 +139,29 @@ build.photo = function (data) {
...
 		thumbnail = `<span class="thumbimg${isVideo ? ' video' : ''}">`;
-		thumbnail += `<img src='${data.thumbUrl}' ` + thumb2x + ` alt='Photo thumbnail' data-overlay='false' draggable='false'>`;
+		thumbnail += `<img class="lazyload" src="dist/images/placeholder.png" data-src='${data.thumbUrl}' ` + thumb2x + ` alt='Photo thumbnail' data-overlay='false' draggable='false'>`;
 		thumbnail += `</span>`
 	} else {
 
 		if (data.small !== '') {
 			if (data.small2x && data.small2x !== '') {
-				thumb2x = `srcset='${data.small} ${parseInt(data.small_dim, 10)}w, ${data.small2x} ${parseInt(data.small2x_dim, 10)}w'`
+				thumb2x = `class="lazyload" src="dist/images/placeholder.png" data-srcset='${data.small} ${parseInt(data.small_dim, 10)}w, ${data.small2x} ${parseInt(data.small2x_dim, 10)}w'`
```
ildyria: I don't think you need the `class="lazyload"` here because it is added in line 156. :)

scripts/main/build.js
```diff
@@ -139,29 +139,29 @@ build.photo = function (data) {
...
 			}
 
 			thumbnail = `<span class="thumbimg${isVideo ? ' video' : ''}">`;
-			thumbnail += `<img src='${data.small}' ` + thumb2x + ` alt='Photo thumbnail' data-overlay='false' draggable='false'>`;
+			thumbnail += `<img class="lazyload" src="dist/images/placeholder.png" data-src='${data.small}' ` + thumb2x + ` alt='Photo thumbnail' data-overlay='false' draggable='false'>`;
 			thumbnail += `</span>`
 		} else if (data.medium !== '') {
 			if (data.medium2x && data.medium2x !== '') {
-				thumb2x = `srcset='${data.medium} ${parseInt(data.medium_dim, 10)}w, ${data.medium2x} ${parseInt(data.medium2x_dim, 10)}w'`
+				thumb2x = `class="lazyload" src="dist/images/placeholder.png" data-srcset='${data.medium} ${parseInt(data.medium_dim, 10)}w, ${data.medium2x} ${parseInt(data.medium2x_dim, 10)}w'`
```
ildyria: I don't think you need the `class="lazyload"` here because it is added in line 164. :)

scripts/main/build.js
```diff
@@ -176,11 +176,11 @@ build.photo = function (data) {
 			}
 
 			if (thumb2x !== '') {
-				thumb2x = `srcset='${data.thumbUrl} 200w, ${thumb2x} 400w'`
+				thumb2x = `class="lazyload" src="dist/images/placeholder.png" data-srcset='${data.thumbUrl} 200w, ${thumb2x} 400w'`
```
ildyria: I don't think you need the `class="lazyload"` here because it is added in line 183. :)

scripts/main/build.js
```diff
@@ -264,7 +264,7 @@ build.imageview = function (data, visibleControls) {
 			let medium = '';
 
 			if (data.medium2x && data.medium2x !== '') {
-				medium = `srcset='${data.medium} ${parseInt(data.medium_dim, 10)}w, ${data.medium2x} ${parseInt(data.medium2x_dim, 10)}w'`;
+				medium = `class="lazyload" src="dist/images/placeholder.png" data-srcset='${data.medium} ${parseInt(data.medium_dim, 10)}w, ${data.medium2x} ${parseInt(data.medium2x_dim, 10)}w'`;
```
ildyria: You don't need the class lazyload here because you are in the image view. As such there is only one picture to load.

alex-phillips: Got it - I completely removed the lazy loading here since there is no point if it's only 1 image

kamil4:
Can you hold on with merging this PR? I want to have a closer look as well, hopefully later today. From a cursory look it seems like there may be duplicate `src` attributes when `2x` is used.

PR #72, opened 2019-04-15
Fix v3 fallback code for thumb2x

kamil4:
Scope of the variables was too limited so the fallback code in thumb2x handling for v3 didn't work.

Also, there was some (harmless) code duplication.

### DIFF
--- a/scripts/main/build.js
+++ b/scripts/main/build.js
@@ -125,7 +125,7 @@ build.photo = function (data) {
 
 	let html = '';
 	let thumbnail = '';
-	let thumb2x = '';
+	var thumb2x = '';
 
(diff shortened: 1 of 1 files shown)


kamil4:
Well, as you know I'm hardly an expert on JavaScript, so I feel like I should be asking _you_ why it makes a difference :wink:. But it does. My guess is that with `let`, the code below basically declares a new variable with a scope that ends with the closing brace. So even though `lychee.retinize` runs and returns correct value, that value is not used for anything. The `thumb2x` variable defined at the top of the method as an empty string never changes.

```
{ // Lychee v3
    let {path: thumb2x} = lychee.retinize(data.thumbUrl)
}
```

PR #78, opened 2019-04-16
fixup default upload import path

ttys3:
for example, the default import path on server is:
`/app/Lychee-Laravel/public/uploads/import/`

not 

`/app/Lychee-Laravel/uploads/import/`

### DIFF
--- a/scripts/main/upload.js
+++ b/scripts/main/upload.js
@@ -418,7 +418,7 @@ upload.start = {
 		};
 
 		basicModal.show({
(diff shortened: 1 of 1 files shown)


ildyria:
Actually no, this should be fixed on Server side otherwise it will break compatibility for the v3.
https://github.com/LycheeOrg/Lychee-Laravel/blob/master/app/Http/Controllers/SessionController.php#L83

ildyria:
Fixed with https://github.com/LycheeOrg/Lychee-Laravel/pull/229

PR #91, opened 2019-05-01
Assorted, v4-specific move/merge fixes

kamil4:
Fix https://github.com/LycheeOrg/Lychee-Laravel/issues/227: allow merging with the parent album.

Display albums and shared albums in the pull-down navigation bar even if only one is present (in v3 if only one was present then it was guaranteed to be us but in v4 that's not the case).

Allow moving/merging with shared albums if admin (admin should be able to do everything, right?).

Add 'Move Album' button to album view (we had one in photo view so it was conspicuously missing here).

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "gulp-rimraf": "^0.2.1",
     "gulp-sass": "^4.0.1",
     "gulp-uglify": "^2.1.1",
-    "jquery": "^3.2.0",
+    "jquery": "^3.4.0",
     "justified-layout": "^2.1.1",
     "lazysizes": "^4.1.7",
     "mousetrap": "^1.6.0",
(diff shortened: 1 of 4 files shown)


kamil4:
I don't get why Travis CI failed -- `npm run compile` works just fine on my system? :confused:

kamil4:
One more thing: for shared albums, the subalbums are not displayed in move/merge context menu. I'm guessing that's because the server doesn't fill them in `albums.json.shared_albums`, unlike with `albums.json.albums` where they are all available at the top level if a user is logged in. Do we need to fix that on the server side so that shared albums behave the same way, at least for the admin?

ildyria:
#button_move_album does not seem to exist right?

kamil4:
The button will magically appear once https://github.com/LycheeOrg/Lychee-Laravel/pull/228 is merged. Oh, I see you're on it already :smile:.

PR #93, opened 2019-05-05
Front end fixes for Lychee-Laravel#223

kamil4:
Fix https://github.com/LycheeOrg/Lychee-Laravel/issues/223.

Fix thumbs of low-resolution videos. The detection of `2x` attributes in json objects was not robust. There was a typo `thumb` vs `thumbUrl`. The fallback for video thumbs was missing. The play arrow should only be displayed once.

Also remove the headers above subalbums and photos when the last photo is deleted (we were already doing so when deleting the last subalbum).

### DIFF
--- a/scripts/main/build.js
+++ b/scripts/main/build.js
@@ -49,7 +49,7 @@ build.getAlbumThumb = function (data, i) {
 	}
 
 	thumb2x = '';
-	if (data.thumbs2x) {
+	if (data.hasOwnProperty('thumbs2x')) {
 		if (data.thumbs2x[i]) {
 			thumb2x = data.thumbs2x[i]
 		}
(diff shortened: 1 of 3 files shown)


scripts/main/build.js
```diff
@@ -128,11 +128,11 @@ build.photo = function (data) {
 	var thumb2x = '';
 
 	let isVideo = data.type && data.type.indexOf('video') > -1;
-	if (data.thumb === 'uploads/thumb/' && isVideo) {
+	if (data.thumbUrl === 'uploads/thumb/' && isVideo) {
 		thumbnail = `<span class="thumbimg"><img src='dist/play-icon.png' alt='Photo thumbnail' data-overlay='false' draggable='false'></span>`
 	} else if (lychee.layout === '0') {
 
-		if (data.thumb2x) { // Lychee v4
+		if (data.hasOwnProperty('thumb2x')) { // Lychee v4
```
d7415: What's the advantage of this / the case it's trying to cover?

As it stands this breaks v3.

kamil4: Oh, I thought I tested with v3 :-(. Let me check and get back to you.

Answering your first question, the test was basically wrong. If `data.thumb2x` existed but was an empty string (s normal circumstance if there is no `thumb2x`) the test would basically resolve as `false` and the v3 fallback code would be executed, even with the v4 back end.

kamil4: OK, I tested. I couldn't get an error in the line you indicated, but I did get it in line 52, which affects smart albums. That change was wrong so I reverted it. I just submitted a fix in #97 -- could you test?

Sorry about that. I should've fixed more thoroughly, and I thought I did, but facts clearly prove me wrong :disappointed:.

d7415: Sorry, looking again it was 52 that broke for me, but I figure if one did there's a reasonable chance the others will too (though I didn't get as far as testing each)

I see what you mean about the test being wrong, but it looks like a list rather than string when set - is it a string when unset? (Yes, I could look, but...)

Just about to look at #97.

kamil4: The problem should be limited to line 52. `thumb2x` is meant to be a string and that part of the fix in #93 should be correct. My mistake was also "fixing" the `thumbs2x` (note the _s_) in line 52, which is indeed not a string but a list, and does not come straight from the server the way `album[s].json` does, but is instead generated in `albums._createSmartAlbums` and is always present, even in v3 (but is `null` then).

PR #97, opened 2019-05-07
Revert an erroneous change from #93

kamil4:


scripts/main/build.js
```diff
@@ -49,7 +49,7 @@ build.getAlbumThumb = function (data, i) {
 	}
 
 	thumb2x = '';
-	if (data.hasOwnProperty('thumbs2x')) {
```
d7415: Looks fine, but just to check, it *is* `'thumbs2x'`, not `'thumb2x'`? Because the others in #93 were `'thumb2x'`...

kamil4: Yes, as I commented in #93, I mixed two issues there. I fixed `thumb2x` detection, which was buggy, but I also accidentally broke `thumbs2x`, which needed no fixing. I guess I searched the front end source for `2x` and changed everything I found. I just went through #93 again and I believe that the remaining changes there should be fine -- only line 52 needed to be reverted.

d7415: Ok. A very quick test of `if (data.hasOwnProperty('thumb2x')) {` on that line seems to work on v3 and v4, but it's your PR and you're the one who's actually looked!

d7415: Read your explanation in #93. Looks good.

kamil4: I instrumented it and also checked with the debugger and I only see `thumbs2x` being present there, not `thumb2x`, and that is the behavior I expect.

Anyway, I'm going to merge this one now as it seems to resolve the problem.

PR #99, opened 2019-05-18
Support new album password unlock functionality

kamil4:
v4-specific patch corresponding to https://github.com/LycheeOrg/Lychee-Laravel/pull/255. It enables the display of thumbs for password-protected albums after the correct password has been provided.


scripts/main/albums.js
```diff
@@ -55,14 +55,10 @@ albums.load = function() {
 
 albums.parse = function(album) {
 
-	if (album.password==='1' && lychee.publicMode===true) {
```
d7415: You're going to make me go think about `publicMode` now, aren't you...

d7415: Doesn't seem to be important (here)

kamil4: `publicMode` tests are generally insufficient in v4, where in addition to v3-style public and logged in (Admin in v4) modes, there are also non-admin users, album sharing between users, and accounts with upload capability disabled. I'm planning to address this issue comprehensively in a future PR but for now I do it case-by-case if it happens to affect the code I work on anyway. In this case, as you say, the `publicMode` case is not required.

kamil4:
@ildyria I think I understand your overall point but I don't understand that https://github.com/LycheeOrg/Lychee-front/blob/master/gulpfile.js#L82. Is this particular example actually used? I don't have `dist/images` in my installation...

ildyria:
Okay, I will fix that.

PR #103, opened 2019-05-21
Fix album selection in sharing.delete

kamil4:
Add additional sanity checks in Users and Sharing views to prevent server errors.

Improve the styling in the Users view (at least I think I improved it :smiley:).

### DIFF
--- a/scripts/main/sharing.js
+++ b/scripts/main/sharing.js
@@ -21,6 +21,15 @@ sharing.add = function () {
 		params.UserIDs += this.value;
 	});
 
+	if (params.albumIDs === '') {
+		loadingBar.show('error', 'Select an album to share!');
+		return false;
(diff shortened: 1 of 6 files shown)


kamil4:
Yeah, the first version of this patch had those corners round, but then I noticed that it was customary in Lychee to have only the bottom corners of buttons be round -- see, e.g., the Settings or Sharing views. So I decided to follow the established convention.

I'll ask @SerenaButler what he thinks :smiley:.

kamil4:
OK, with my latest push, the top corners should be consistently round in all the views. And yes, I kept them at the right angle for the inner corners in Save/Delete.

Also, I forgot to mention that this commit should fix https://github.com/LycheeOrg/Lychee-Laravel/issues/143.

PR #106, opened 2019-05-26
Prevent attempts to edit shared albums by others

kamil4:
Work in progress. It works for me but needs a little more testing (in particular, I still need to test if I didn't break anything with v3). Some input wouldn't hurt either :smiley:.

The primary functionality is the disabling of editing capabilities in shared albums for regular users. Given that we can trigger edits using header buttons, context menu, and keyboard bindings, it's a lot of code to tweak... Once this is merged, https://github.com/LycheeOrg/Lychee-Laravel/issues/219 can be closed... Also, this part depends on https://github.com/LycheeOrg/Lychee-Laravel/pull/260.

As I was implementing this, I got distracted by something that was on my todo list for a while: enabling the pull-down album/photo list in `publicMode`. Until now it was only available for logged in users, for no good reason as far as I can tell. I also added the list to the albums view (it may not be all that useful there in v3, but with nested albums in v4, it's a different story).

Questions/comments:

I added a new `album.isUploadable()` function which replaces virtually all tests for `lychee.publicMode`. Looking back at how it's used, I'm no longer sure if it should be under `album` or under `lychee` -- as it can safely be invoked when `album.json === null`, and in fact there are many such invocations in the code. So as an exception I welcome bikeshedding input regarding the name :wink:.

The pull-down list used to be available only for editable albums/photos, but now it's available for all. Still, there are a number of instances in the code where it's referred to as "editable". I didn't bother to change those -- should I?

### DIFF
--- a/scripts/main/contextMenu.js
+++ b/scripts/main/contextMenu.js
@@ -146,28 +146,29 @@ contextMenu.albumTitle = function(albumID, e) {
 
 		let items = [];
 
-		items = items.concat({ title: lychee.locale['ROOT'], fn: () => lychee.goto()});
+		items = items.concat({ title: lychee.locale['ROOT'], disabled: (albumID === false), fn: () => lychee.goto()});
 
 		if (data.albums && data.albums.length > 0) {
 
 			items = items.concat({});
-			items = items.concat(contextMenu.buildList(data.albums, [ parseInt(albumID, 10) ], (a) => lychee.goto(a.id)));
+			items = items.concat(contextMenu.buildList(data.albums, (albumID !== false ? [ parseInt(albumID, 10) ] : []), (a) => lychee.goto(a.id)));
 
 		}
 
 		if (data.shared_albums && data.shared_albums.length > 0) {
 
 			items = items.concat({});
-			items = items.concat(contextMenu.buildList(data.shared_albums, [ parseInt(albumID,10) ], (a) => lychee.goto(a.id)));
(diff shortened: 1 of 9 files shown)


scripts/main/contextMenu.js
```diff
@@ -418,11 +423,13 @@ contextMenu.shareAlbum = function(albumID, e) {
 		{ title: build.iconic('facebook', iconClass) + 'Facebook', fn: () => album.share('facebook') },
 		{ title: build.iconic('envelope-closed') + 'Mail', fn: () => album.share('mail') },
 		{ },
-		{ title: build.iconic('pencil') + lychee.locale['EDIT_SHARING'], visible: lychee.publicMode===false, fn: () => album.setPublic(albumID, true, e) },
-		{ title: build.iconic('ban') + lychee.locale['MAKE_PRIVATE'], visible: lychee.publicMode===false, fn: () => album.setPublic(albumID, false) }
```
d7415: Just to confirm, we don't need these `visible` restrictions because we're removing the option altogether in the splice?

kamil4: Exactly. It was redundant.

ildyria:
Looks good to me.

ildyria:
Tested on Lychee vanilla. I did not notice wrong behavior.

kamil4:
I added a (trivial) commit to correctly handle missing video thumbnails (not a bug I introduced, but it's more obvious now).

PR #101, opened 2019-05-20
Public search functionality

copperschnack:
Hi Lychee-team,

as needed by us, and demanded in https://github.com/LycheeOrg/Lychee-Laravel/issues/49, we implemented a public search functionality in Lychee. Please feel free to review and maybe integrate it into your main branch.

As this feature comes with some changes to the backend, we will file a corresponding pull request there.

This is our first public pull request - so feedback is highly appreciated.

Thanks and regards

malte.

PR #110, opened 2019-05-29
Public search

kamil4:
Followup to #101, this time on a branch we can all easily push to...


kamil4:
OK, this is now tested and should be ready to go both with v3 and v4.

ildyria:
Rebase on master (was the first `push --force`).
Rewrite the history with the correct names (was the second `push --force`)...

Issue #84, opened 2019-04-18
No "unselecting" possible


SerenaButler:
When select items with rubberband no "unselecting" is possible, e.g. click in empty space and no "reselecting" is possible, using rubberband again. Only a click on a image, going in image view with the need to going back to album view is working.

![image](https://user-images.githubusercontent.com/35893306/56357155-23ed5800-61db-11e9-9cfb-2daf11b36f56.png)

SerenaButler:
AFTER the context menu was in use unselecting by clicking on the background IS possible ;-)

kamil4:
This one's done -- #116 will fix it when merged (hint, hint :wink:).

kamil4:
Closed via #116.

SerenaButler:
Not fixed, sorry .Still not working :-(

kamil4:
Did you recompile the front end (`npm run compile`)? The precompiled version in the `Lychee-Laravel` repo does not include those fixes yet.

SerenaButler:
Ok that might be the reason. But can't do that on my hosting. I wait then ;-)

kamil4:
OK, try again -- @ildyria updated the generated files as part of his recent big merge.

SerenaButler:
Working! :-)

PR #116, opened 2019-06-15
Assorted multiselect fixes

### LINKED ISSUE #75: Fix shift + click
When select an image, shift click on a second image, all images between these two should be selected to mirror usual UX.

kamil4:
Fixes #75, #84, and more.

Rational support for "Select all" in albums with photos and subalbums
(do not generate an error message).

Add support for Shift key to select a range of albums or photos.

Intuitive clearing of selection with a single click.

multiselect.removeItem() didn't work correctly for non-first items.

Selection couldn't start much below the last picture in albums with just
a few pictures (i.e., those that had whitespace there).

Cleaned up calculations of selection and overlap.

New multiselect.unbind() function to match the existing
multiselect.bind().

### DIFF
--- a/scripts/main/multiselect.js
+++ b/scripts/main/multiselect.js
@@ -29,7 +29,7 @@ multiselect.position = {
 
 multiselect.bind = function() {
 
-	$(document).on('mousedown', (e) => { if (e.which===1) multiselect.show(e) });
+	$('.content').on('mousedown', (e) => { if (e.which===1) multiselect.show(e) });
 
 	return true
 
@@ -37,7 +37,7 @@ multiselect.bind = function() {
 
 multiselect.unbind = function() {
 
-	$(document).off('mousedown');
+	$('.content').off('mousedown');
 
 };
 
@@ -256,7 +256,7 @@ multiselect.show = function(e) {
 		multiselect.clearSelection()
 	}
 
-	multiselect.position.top    = e.pageY+1;
+	multiselect.position.top    = e.pageY;
 	multiselect.position.right  = $(document).width() - e.pageX;
 	multiselect.position.bottom = $(window).height() - e.pageY;
 	multiselect.position.left   = e.pageX;
--- a/scripts/main/multiselect.js
+++ b/scripts/main/multiselect.js
@@ -256,7 +256,7 @@ multiselect.show = function(e) {
 		multiselect.clearSelection()
 	}
 
-	multiselect.position.top    = e.pageY;
+	multiselect.position.top    = e.pageY+1;
 	multiselect.position.right  = $(document).width() - e.pageX;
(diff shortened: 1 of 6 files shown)


kamil4:
I changed the behavior of Shift-click to be as we discussed on Gitter. This should be ready to go in now.

d7415:
> **REGRESSION: the button "<" (back) in albums and sub albums is not working anymore.**

I guess it's treating it as a select - if you click enough times in quick succession it does still work

> Aside from that I love the new behaviour.

Agreed

scripts/main/multiselect.js
```diff
@@ -256,7 +256,7 @@ multiselect.show = function(e) {
 		multiselect.clearSelection()
 	}
 
-	multiselect.position.top    = e.pageY;
+	multiselect.position.top    = e.pageY+1;
```
d7415: This is a bit of a hack, but seems to fix the back button. Better solutions welcome!

kamil4:
Ugh :-(. This regression is due to a change I made to `mousedown` events: I bound them to the whole document rather than to `content` (as it used to be) because the former covers the whole window whereas the latter for nearly empty albums leaves a gap between the content and the footer where multiselect cannot be initiated.

I didn't realize it would cause literally *all* mouse-click events to go through our whacky multiselect code. I hate that. A weird thing is that it seems to mostly work for me in Firefox (which is why I never noticed this) but it does fail spectacularly in Chrome. Even with @d7415's tweak (and to be completely honest I don't understand how it works, so respect :wink:) I can still produce other weird behavior, e.g., by opening the "+" menu and clicking outside of it to close it, a multiselect is initiated.

I'm going to revert the change to `mousedown`. It will be much safer that way. Let me know if it works for you.

Is there a way though to extend the `content` divider to the very edge of the footnote even if there is not enough content to fill it? That would ensure that any click in that area can initiate a multiselect.

d7415:
> Is there a way though to extend the `content` divider to the very edge of the footnote even if there is not enough content to fill it? That would ensure that any click in that area can initiate a multiselect.

I don't have the time to look right now, but that certainly seems the right way to do this (unlike my hack, but I did advertise it as such!)

kamil4:
Well, I remembered that @ildyria recently fought with the height of the `content` container and had to revert. Sure enough, I encountered the same problem of the vertical spacing between albums getting stretched out. I was able to track it down and turn it off though.

In the process I also noticed that with a sufficiently short window, the footer will overlap the sidebar. I fixed that as well.

This should really be ready to go now but I would appreciate you trying it out again as I have noticed that any tweaks of element dimensions or margins we've recently done tended to require follow-up patches to fix unexpected collateral damage. So I do need others to try it out...

ildyria:
I fixed a small thing that was annoying me. Other than that, looks good to go for me.

PR #127, opened 2019-07-03
Add Download to context menus

kamil4:
Fixes https://github.com/LycheeOrg/Lychee-Laravel/issues/216.

Support for download via multiselect is v4-only...

Note that since this depends on modifications to the server protocol, the corresponding server changes (https://github.com/LycheeOrg/Lychee-Laravel/pull/296) need to be merged at the same time.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -681,15 +681,15 @@ album.share = function (service) {
 
 };
 
-album.getArchive = function (albumID) {
+album.getArchive = function (albumIDs) {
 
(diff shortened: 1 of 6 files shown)


kamil4:
I pushed additional changes to fix the view mode (I still plan to kill it, but all in due time...). It requires corresponding server changes both to v3 and v4.

v4 server now also adds an additional `downloadable` field to the `Photo::get` response to make it easy for the front end to figure out whether to provide the download menu items or not.

PR #130, opened 2019-07-15
Nextprev improvements

### LINKED ISSUE #129: [Enhancement] Display of images within lightbox (preload and use of low-res)
_Explain in a few words which functionality or improvements you would like to see in Lychee._

When having a slow connection, the display of images within the lightbox is not optimal. This can for example happen on your mobile if 4G is not available or the connection is consumed by some other application.

**Observed behavior:**
* Lightbox is opened, but image is missing
* Image gets loaded and builds up slowly
(move the next image)
* Image is missing, gets loaded and builds up slowly

**I'd like to propose two enhancements to improve the UX:**
(1) Use low-res to start with
- Use the already cached low resolution version of the album view to start with
- Once the actual image for the lightbox is loaded, change to this version

(2) Pre-load of images
- Preload the neighboring (i.e. predecessor and successor) images of the album when a new image is display in the lightbox

kamil4:
Fixes #129.

The display of the thumbnail in the photo view while the larger-resolution image is being loaded is implemented using CSS background image (on the `` object, no less). It actually looks pretty cool, I think.

Image prefetching code was extended to request both previous and next images. The prefetch request is initiated only after the main image has finished loading. We no longer prefetch movie files. There's now a fallback for browsers that don't support HTML5 `link rel=prefetch`. While the fallback was trivial, detecting whether it's needed was not. I thought I would be forced to hardcode it based on the user agent, but in the end I found some code that's supposed to detect it at run time.

Code was also extended in places to handle the case of previous or next image being empty, to fully support non-wraparound feature being added to the server.

### DIFF
--- a/scripts/main/photo.js
+++ b/scripts/main/photo.js
@@ -67,8 +67,7 @@ photo.load = function(photoID, albumID) {
 		lychee.imageview.show();
 
 		setTimeout(() => {
-			lychee.content.show();
-			photo.preloadNextPrev(photoID)
+			lychee.content.show()
 		}, 300)
 
 	})
@@ -170,9 +169,9 @@ photo.preloadNextPrev = function(photoID) {
 						href = preloadPhoto.medium2x
 					}
 				}
-			} else if (preloadPhoto.type && data.type.indexOf('video') === -1) {
+			} else if (preloadPhoto.type && preloadPhoto.type.indexOf('video') === -1) {
 				// Preload the original size, but only if it's not a video
-				href = url
+				href = preloadPhoto.url
 			}
 
 			if (photo.supportsPrefetch === null) {
--- a/scripts/main/photo.js
+++ b/scripts/main/photo.js
@@ -4,8 +4,9 @@
 
 photo = {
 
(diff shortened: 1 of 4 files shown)


scripts/main/photo.js
```diff
@@ -139,33 +139,75 @@ photo.update_display_overlay = function () {
...
+
+				if (preloadPhoto.medium2x && preloadPhoto.medium2x !== '') {
+					if (current2x === null) {
+						let imgs = $('img#image');
+						current2x = imgs.length > 0 && imgs[0].currentSrc !== null && imgs[0].currentSrc.includes('@2x.')
+					}
+					if (current2x) {
+						// If the currently displayed image uses the 2x variant,
+						// chances are that so will the next one.
+						href = preloadPhoto.medium2x
```
ildyria: Yes and no. I think you still need to check if medium2x is not empty.

d7415: Is that not covered on line 161?

ildyria: Ho right. xD

PR #133, opened 2019-08-01
Make content area shorter

d7415:
...so that the v4 footer doesn't overlap the bottom of the page.

### DIFF
--- a/scripts/main/lychee.js
+++ b/scripts/main/lychee.js
@@ -94,6 +94,8 @@ lychee.aboutDialog = function() {
 
 lychee.init = function() {
 
+	lychee.adjustContentHeight();
+
 	api.post('Session::init', {}, function(data) {
 
 		lychee.api_V2 = data.api_V2 || false;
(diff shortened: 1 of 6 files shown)


ildyria:
The problem with that one is when someone only have one thing enabled in the footer like just the copyright but not the social icons and has only one row of albums. :/

kamil4:
Sorry, I'm responsible for the `min-height: calc(100vh - 83px - 36px);` in `_content.scss`. I only have a single line in my footers and I forgot that one can have more (and that two lines is in fact the default).

I just played with the settings and I managed to get the footer up to 126px tall (with the social icons, the copyright line, and an additional line all enabled). So your patch is not sufficiently conservative...

But yeah, as @ildyria pointed out, there is no one correct value that we could hardcode into the CSS that will look right in all circumstances. I think we will have to do it dynamically, from JS. I believe that the `min-height` line for `div#container` is not needed at all at this point and can be commented out. The one that matters is `min-height` for `.content`. I'm guessing we can query from JS the height of the `div#footer` DOM object after building it and then dynamically adjust `.content`'s CSS?

ildyria:
> I'm guessing we can query from JS the height of the `div#footer` DOM object after building it and then dynamically adjust `.content`'s CSS?

The problem of that approach is the flickering. :s

kamil4:
> The problem of that approach is the flickering.

I think if we do it early enough we might be fine. Especially since (as I just realized) the footer is supplied by the backend so we don't need to wait for it to get built. Keep in mind that the overall height of the page is also determined dynamically, based on the number of albums/pictures (which are laid out after the rest of the page is built), and we don't notice any flickering there...

d7415:
Ok, so this is less of a quick fix than I'd hoped. If/when I get time I'll look properly, but if anyone wants to take over in the meantime, please do!

kamil4:
FYI, I'm working on this one...

kamil4:
OK, I overhauled this part of the code. Hopefully it will be much more robust now (famous last words :wink:).

Footer now uses absolute positioning so it is firmly bound to the bottom of the page, ensuring no flicker as we adjust other parameters. The trick to make this work, as I found out after some pain, is to set the positioning of `body` to `relative` (otherwise, the footer is bound to the bottom of the initially visible portion of the page and scrolls up with it).

Speaking of which, the height of `html` and `body` are no longer set to `100%`, which I believe was a mistake. Instead, `min-height` is set to `100vh` to ensure that the footer is indeed always at the bottom while `height` is allowed to grow naturally with the content of the page.

To ensure no overlap between the content and the footer, the padding of the container around the content is set to the height of the footer. `min-height` of the content takes the height of the footer into account as well. This adjustment is made early on from JS, once, during `lychee.init`, before any content is loaded.

ildyria:
Good for merge. :)

PR #134, opened 2019-08-15
Sharing

kamil4:
* New share icon and a copy to clipboard icon
* Split visibility and sharing icons in album and photo views
* Rework album sharing to have all controls in a single dialog box; support preserving of the current password (requires server-side changes)
* Get rid of "hidden" which wasn't there in v3 and is redundant
* Use green and orange for the visibility icon, as appropriate
* Add a photo sharing dialog box to match album sharing
* Support the download of all image sizes (v4 only; requires server-side changes)
* Add a copy-to-clipboard option for albums and a dialog box for all image sizes; get rid of the read-only input box
* Make loading bar notifications easier to see when a dialog window is open
* Fix individual photo sharing display in the info sidebar
* Change the badge colors of the Public smart album and individually shared photos to orange
* Fix the most recent breakage in the standalone view
* Add support for sliders and disabled checkboxes to basicModal.


ildyria:
Code wise it looks good to me.
However for some reason the Copy to Clipboard function does not work here (Debian with Chrome 76).

kamil4:
@ildyria Are you sure? I just tested with 76.0.3809.87 under Linux and copy to clipboard works fine here... Are you getting some error or the expected "OK: Copied URL to clipboard!" notification?

On a somewhat unrelated note, I notice that the first link in _Direct Links_ (in photo view) is selected in Chrome, which I went to some pain to avoid, but apparently I was only successful with that goal under Firefox. I'll see if I can get that figured out...

PR #136, opened 2019-08-31
Basiccontext

kamil4:
* Incorporate the sources of basiccontext
* Remove the unnecessary 'overflow: hidden' which can cause reflow
* Also use the minimized version of lazysizes


ildyria:
I would be in favor to put this not in `script` and `style` but in another folder like `deps/scripts` and `deps/style` to separate the code of the front from the *dependencies*.

kamil4:
@ildyria No problem. I asked a while back where to put them but, given that I didn't get a response, I figured I would force the issue by submitting a PR. Clearly it worked :wink:.

@d7415 Thanks! I figured there would be no problem given that they came from the same person but it's good to know that you checked the licenses...

PR #139, opened 2019-09-10
Close lightbox by swiping down the photo (mobile only)

### LINKED ISSUE #112: [Enhancement] Swipe down to close Lightbox
To increase the mobile UX, it would be nice to be able to swipe down to close the lightbox. This behavior is well known to users on mobile devices.

It seems like the jquery swipe event only covers horizontal swiping. But there are a couple of libraries which provide support for vertical swiping, e.g. 
https://github.com/hammerjs/hammer.js
https://github.com/mattbryson/TouchSwipe-Jquery-Plugin

tmp-hallenser:
Fixes #112.

Background stays solid, future enhancement can be to make it dynamically transparent.

### DIFF
--- a/scripts/main/init.js
+++ b/scripts/main/init.js
@@ -101,7 +101,7 @@ $(document).ready(function() {
 
 			// Fullscreen on mobile
 			.on('touchend', '#imageview #image', function(e) {
-				if (swipe.obj==null || (swipe.offset>=-5&&swipe.offset<=5)) {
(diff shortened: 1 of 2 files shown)


kamil4:
I tested your PR on Android under Firefox and Chrome and it works fine. The source code does not have any surprises either. @ildyria, can you have a look as well?

So the lightbox can be closed not only by swiping down but also by swiping up. I'm assuming that's intentional?

PR #148, opened 2019-09-24
Applied translations

joserick:
Hi, 

I saw that several translations do not work, so I added this change to be displayed.

But I do not apply all the translations because multiple words like "More", "Full configuration", "Search" among others that do not exist in "php / Locale / English.php" and I would like to add them but I don't know if I also have to add them to "Lychee-front / scripts / main / lychee_locale.js"? and I also don't know if there would be a problem with this new translation in the other languages? (Italian.php, German.php, etc).

Could you help me with these questions before changing anything.

Thank you.

### DIFF
--- a/scripts/main/view.js
+++ b/scripts/main/view.js
@@ -11,6 +11,8 @@ view.albums = {
 		multiselect.clearSelection();
 
 		view.albums.title();
+		view.albums.setting();
+		view.albums.add();
 		view.albums.content.init();
 
 	},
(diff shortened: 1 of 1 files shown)


kamil4:
So why did you close this PR? Are you going to solve the problem in a different way?

While I'm not thrilled about adding that many new functions, I can certainly understand the problem you are trying to solve. For Lychee-Laravel (the upcoming Lychee v4) at least, I'm guessing you should be able to put the localization strings directly in the blades (see, e.g., https://github.com/LycheeOrg/Lychee-Laravel/blob/master/resources/views/gallery.blade.php) and that would certainly be our preferred approach to fixing this problem. But there are probably also hardcoded strings in the front end that need to be localized as well.

PR #149, opened 2019-09-24
Added function to apply translations to the header

joserick:
Hi,

I saw that several translations do not work, so I added this change to be displayed.

But I do not apply all the translations because multiple words like "More", "Full configuration", "Search" among others that do not exist in "php / Locale / English.php" and I would like to add them but I don't know if I also have to add them to "Lychee-front / scripts / main / lychee_locale.js"? and I also don't know if there would be a problem with this new translation in the other languages? (Italian.php, German.php, etc).

Could you help me with these questions before changing anything.

Thank you.

### DIFF
--- a/scripts/main/header.js
+++ b/scripts/main/header.js
@@ -232,7 +232,7 @@ header.setEditable = function(editable) {
 
 };
 
-header.applyTransduction = function () {
+header.applyTranslations = function () {
 
 	let selector_locale = {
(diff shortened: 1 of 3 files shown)


joserick:
Hi,

Sorry @kamil4, I had not read your answer, Yes, I closed ( #148 ) because I had found a better way to apply the translations, if I understand well the Laravel project is the main one now. Then I will check the code and see how I can help.

PR #150, opened 2019-09-26
Small fixes

### LINKED ISSUE #142: Multiple image move between albums
Dear friends, I like this project and I am very sattisfiet with it.
There is one thing, I am missing:
**Possibilty to select more than one image and move these selced images to another existing or new album.**
Should this be a new feature of any future build?
Many thanks for Lychee-Laravel, the best sel.hosted photo album!

kamil4:
Fixes #142.

Assorted minor fixes:
* Next/Prev photo prefetch was misbehaving on Firefox for HiDPI images
* Prefetch was doing more than it should for video files
* An earlier optimization that eliminated unnecessary reloads during image moves didn't handle the case of moving to a new album so it got reverted
* Wrapping of html code with `lychee.html` was added where missing

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -250,7 +250,7 @@ album.add = function (IDs = null, callback = null) {
 	};
 
 	basicModal.show({
-		body: `<p>${lychee.locale['TITLE_NEW_ALBUM']} <input class='text' name='title' type='text' maxlength='50' placeholder='Title' value='Untitled'></p>`,
+		body: lychee.html`<p>${lychee.locale['TITLE_NEW_ALBUM']} <input class='text' name='title' type='text' maxlength='50' placeholder='Title' value='Untitled'></p>`,
 		buttons: {
 			action: {
 				title: lychee.locale['CREATE_ALBUM'],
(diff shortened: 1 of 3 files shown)


kamil4:
My thought process for adding `lychee.html` was pretty simple: while working on something else I realized that most, if not all html fragments in the front end use `lychee.html`, and I remembered that in the code I added recently I did not do that, so I added them. It didn't seem to change anything, so I figured it was fine. No deep thinking involved :wink:.

Now that you called me out on this, I had a look at what `lychee.html` actually does and, frankly, I'm still not sure what the right thing to do is. It seems that the escaping of characters only happens when we use _double_ $ sign in JS (`$$`). We do indeed do that in the front end in some places but not in others; I currently lack intuition on when to use which one. Is it that any user-provided (or -derived) data should be `$$`-escaped but other uses (such as `lychee.locale`) should use a single `$`, to prevent the problem you mention? I didn't use `$$` anywhere; perhaps I should've in some places?

kamil4:
OK, I verified that `lychee.html` does not escape characters in regular (single-`$`) cases but it does do so for `$$`, as expected. So the changes I made should be safe.

PR #152, opened 2019-10-02
Adding functionality to display photos of an album on a map

tmp-hallenser:
This PR adds the functionality to display photos of an album (and its sub albums) on a map.

* Adds 3 new npm packages (spin.js, leaflet-spin, leaflet.markercluster) and an additional leaflet extension (https://github.com/turban/Leaflet.Photo branch gh-pages)
* Map is opened when clicking on map icon within the header (only album and photo view)
* When clicking on a thumbnail within the map, a preview is opened. Clicking on the preview open the normal photo view of the specific photo
* URL changes to #map/albumID
* Map functionality can be deactivated within settings, resulting in a status message when trying to open it


ildyria:
> Map functionality can be deactivated within settings, resulting in a status message when trying to open it

- Is the map icon still visible if the Map functionality if available ?
- Is it possible to disable for the public the Map Icon but not for logged users?

:)

tmp-hallenser:
> > Map functionality can be deactivated within settings, resulting in a status message when trying to open it
> 
> * Is the map icon still visible if the Map functionality if available ?
Map icon is always visible, even if functionality is disabled
> * Is it possible to disable for the public the Map Icon but not for logged users?
Not yet possible (would imho only make sense if you also disable the info panel for photos since you can extract the location from there as well)
> 
> :)

ildyria:
>   Map icon is always visible, even if functionality is disabled

Then you should make it disappear if the functionality is disabled :stuck_out_tongue_closed_eyes: 

> > * Is it possible to disable for the public the Map Icon but not for logged users?
>
> Not yet possible (would imho only make sense if you also disable the info panel for photos since you can extract the location from there as well)

That would be a nice idea to decide whether you make the map info public or not. :)

tmp-hallenser:
> > Map icon is always visible, even if functionality is disabled
> 
> Then you should make it disappear if the functionality is disabled 😝

done
> > > * Is it possible to disable for the public the Map Icon but not for logged users?
> > 
> > 
> > Not yet possible (would imho only make sense if you also disable the info panel for photos since you can extract the location from there as well)
> 
> That would be a nice idea to decide whether you make the map info public or not. :) You may not want to broadcast your house location to the world. ;)

done
> Ideally you would like this to be a per album feature. Have a look at the way `full_photo` is implemented. :)

Let's leave this feature for the future, ok?

Any further suggestions?

There is one minor issue which I almost forgot. The leaflet.markercluster plugin wants the leaflet.markercluster.js.map file to be located unter (home)/dist. Although it's only for debugging - how to I instruct the gulp file to copy it to this location.

kamil4:
Gee, you are a volcano of activity, @tmp-hallenser! :smile:

I haven't tried this new feature yet, but based on a casual look over the source code, I have the following remarks/concerns:

* Another button in the header. Have you tested if they all still fit on mobile (in particular, screens 320px wide)? We were pretty close to full already, from what I remember. It's an issue we'll need to tackle comprehensively as we keep adding more features; for now, we just keep kicking the bucket down the road...
* HiDPI support. When I experimented with the little map in the sidebar that you implemented first, it didn't have HiDPI support for the map data itself, only for the marker. I felt that was OK since, well, it was just a little map in the sidebar. Now that it's a whole separate view, we should strive to provide an optimal-quality rendering, both for the map and for the picture previews.
* The decision to include pictures from the subalbums is... interesting. It is not in line with the rest of the Lychee interface, where we tend to consider each album separately. We make an exception for album downloads though, so there is a precedent... Ideally, I think an "Include subalbums" checkbox would be handy, both here and in other parts of the Lychee interface.

Note that these are just my thoughts and _not _a list of must-implement-before-we-even-consider-merging-it features... :smile:

ildyria:
> There is one minor issue which I almost forgot. The leaflet.markercluster plugin wants the leaflet.markercluster.js.map file to be located unter (home)/dist. Although it's only for debugging - how to I instruct the gulp file to copy it to this location.

Have a look at this: https://github.com/LycheeOrg/Lychee-front/blob/master/gulpfile.js#L315

You can create a task for this copy and add it near https://github.com/LycheeOrg/Lychee-front/blob/master/gulpfile.js#L345

tmp-hallenser:
I’ve submitted several commits covering the following:
A. HiDPI Support
* Map now has HiDPI Support, e.g. map tiles, map markers, thumbnails and previews (as far as HiDPI Source is available)
* Switch of default map tile provider to Wikimedia providing HiDPI tiles
B. Include pictures of sub albums
* Added config switch to control this behavior
C. Further changes
* Added switch to change map tile provider
* Added map at front page
* Refactored calls to backend for location data (now only a single call)
* Fixed issue with leaflet.Markercluster (map.js file)

ildyria:
This looks pretty good and this does not impact Lychee without the map module.

PR #153, opened 2019-10-04
Added basic support for uploading images by pasting from clipboard

Skayo:
### Description:
This PR adds support for uploading images by pasting them from clipboard into Lychee.
Tested in Chrome 77 and Firefox 69 on Windows 10.

**Important:** This does not add support for uploading images by pasting an **URL** from clipboard, as requested in the Feature Request.
That's because I found it unnecessary to have this feature when you could just use the normal "+" -\> "Import from Link" method

**Also Important:** On Windows 10, I've tried to copy an image file in Explorer and paste it into Lychee. This didn't work for some reason. But what definitely works is going to a random website, clicking on an image, clicking "Copy Image" and then pasting it into Lychee!

### Example Use:
1. Go on a random website with images (for example [unsplash.com](https://unsplash.com))
2. Click on an image you want to import to Lychee
3. Right-Click on the image and choose "Copy Image"
4. Go to Lychee
5. Press "Strg+V"

### DIFF
--- a/scripts/main/init.js
+++ b/scripts/main/init.js
@@ -154,29 +154,29 @@ $(document).ready(function() {
 	// Paste upload
 	.on('paste', function (e) {
 		if (e.originalEvent.clipboardData.items) {
-			if (!album.isUploadable()) {
-				return false;
+			const items = e.originalEvent.clipboardData.items;
+			let filesToUpload = [];
+
+			// Search clipboard items for an image
+			for (let i = 0; i < items.length; i++) {
+				if (items[i].type.indexOf('image') !== -1 || items[i].type.indexOf('video') !== -1) {
+					filesToUpload.push(items[i].getAsFile());
+				}
 			}
 
-			// Close open overlays or views which are correlating with the upload
-			if (visible.photo()) lychee.goto(album.getID());
-			if (visible.contextMenu()) contextMenu.close();
-
-			if (e.originalEvent.clipboardData.items.length > 0) {
-				const items = e.originalEvent.clipboardData.items;
-				let imagesToUpload = [];
-
-				// Search clipboard items for an image
-				for (let i = 0; i < items.length; i++) {
-					if (items[i].type.indexOf('image') !== -1) {
-						imagesToUpload.push(items[i].getAsFile());
-					}
+			if (filesToUpload.length > 0) {
+				if (!album.isUploadable()) {
+					return false;
 				}
 
-				upload.start.local(imagesToUpload);
-			}
+				// Close open overlays or views which are correlating with the upload
+				if (visible.photo()) lychee.goto(album.getID());
+				if (visible.contextMenu()) contextMenu.close();
 
-			return false;
+				upload.start.local(filesToUpload);
+
+				return false;
+			}
 		}
 	})
 
--- a/scripts/main/init.js
+++ b/scripts/main/init.js
@@ -165,17 +165,25 @@ $(document).ready(function() {
 			}
 
 			if (filesToUpload.length > 0) {
-				if (!album.isUploadable()) {
-					return false;
-				}
+				if (!album.isUploadable())
+					return;
 
 				// Close open overlays or views which are correlating with the upload
-				if (visible.photo()) lychee.goto(album.getID());
-				if (visible.contextMenu()) contextMenu.close();
 
-				upload.start.local(filesToUpload);
+				if (visible.photo())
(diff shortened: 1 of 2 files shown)


scripts/main/init.js
```diff
@@ -151,6 +151,35 @@ $(document).ready(function() {
...
+			// Close open overlays or views which are correlating with the upload
+			if (visible.photo()) lychee.goto(album.getID());
+			if (visible.contextMenu()) contextMenu.close();
+
+			if (e.originalEvent.clipboardData.items.length > 0) {
+				const items = e.originalEvent.clipboardData.items;
+				let imagesToUpload = [];
+
+				// Search clipboard items for an image
+				for (let i = 0; i < items.length; i++) {
```
kamil4: Are you a C programmer by training? :smiley:

`forEach`/`$.each` would be a more JavaScript-y way of iterating over these...

scripts/main/init.js
```diff
@@ -151,6 +151,35 @@ $(document).ready(function() {
...
+					if (items[i].type.indexOf('image') !== -1) {
```
kamil4: Lychee also supports video files, so I guess you might want to test for them as well?

scripts/main/init.js
Skayo: Oh I'm just used to using a for loop. I always have this "It's faster that way" in my mind.
Will change.

scripts/main/init.js
Skayo: Sure I'll check it out!

scripts/main/init.js
Skayo: Okay nevermind it doesn't work with `forEach`/`$.each` because `items` isn't an array but a `DataTransferItemList`

scripts/main/init.js
Skayo: @kamil4 Okay so I can't manage to copy a video to clipboard on Windows 10. Is it possible on Linux?

scripts/main/init.js
kamil4: My bad...

scripts/main/init.js
kamil4: No idea. I'm not sure if I've ever copied an image file over clipboard, let alone a video... But I intend to try when I test your PR this weekend :smile:

d7415:
It's not working on firefox for me, but is in chromium

Skayo:
@kamil4 I caused a little bug that prevents from pasting anything into Lychee except images. So when I for example rename a photo I couldn't paste text into the input field... Fixed this now.

Also did some little movements of the code. Nothing special.

Skayo:
> @d7415: It's not working on firefox for me, but is in chromium
Weird. As written above, it works for me in Firefox 69 on Windows 10...
I'll google around a bit. Maybe I find the reason for this...

d7415:
> > @d7415: It's not working on firefox for me, but is in chromium
>
> Weird. As written above, it works for me in Firefox 69 on Windows 10...
> I'll google around a bit. Maybe I find the reason for this...

It might just be my setup. Looking for second opinions.

kamil4:
@d7415 What is the value of `dom.event.clipboardevents.enabled` in your Firefox? I had mine set to `false`, which was indeed preventing the paste from working...

kamil4:
OK, I tested it and it works. I couldn't copy a video to the clipboard either but I suppose there's no harm in leaving that condition in.

One thing that still worries me is how global this feature currently is. If the clipboard contains an image and I (perhaps accidentally) select Paste, the image will get added to the current album, no matter what I was just doing. Say, even if I was just editing text in a text field, for instance, changing the title of a photo. I think it would be safer if the paste event worked only when `visible.albums() === true` or `visible.album() === true` and was ignored if `basicModal.visible() === true` or `visible.contextMenu() === true` (and perhaps also if `visible.leftMenu() === true`, so that it catches, say, Settings).

Skayo:
@kamil4 Well the drag and drop upload also works on the whole site with the same conditions as the paste upload. But sure I can change it!

d7415:
> @d7415 What is the value of `dom.event.clipboardevents.enabled` in your Firefox? I had mine set to `false`, which was indeed preventing the paste from working...
> 
> I think I had it that way because of this: https://www.howtogeek.com/251807/how-to-enable-pasting-text-on-sites-that-block-it/

I'm sure I checked about:config.... Maybe I mistyped "clipboard" :P Good spot!

PR #156, opened 2019-10-12
Fix 356

ildyria:
Add link for update in the left, separate the diagnostic answer from the update check.

### DIFF
--- a/scripts/main/view.js
+++ b/scripts/main/view.js
@@ -1566,21 +1566,14 @@ view.diagnostics = {
 		lychee.content.html(html);
 
 		$("#Update_Lychee").on('click', function () {
-			api.get('api/Update', function (data) {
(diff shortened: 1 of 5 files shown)


scripts/main/view.js
```diff
@@ -1629,3 +1632,53 @@ view.diagnostics = {
...
+};
+
+view.preify = function (data, css) {
+	html = '<pre class="' + css + '">';
+	if (Array.isArray(data)) {
+		for (let i = 0; i < data.length; i++) {
+			html += '    ' + data[i] + '\n';
+		}
+	} else {
+		html += '    ' + data_json;
```
kamil4: Isn't `data_json` going to be undefined here?

ildyria: oups ^^

Issue #49, opened 2019-03-11
Album View // About // Share Settings


SerenaButler:
It makes no sense to show that when not logged in ;-)

It's obviously public, not hidden nor password protected ... :-P And the owner ... I wrote about :-D

![image](https://user-images.githubusercontent.com/35893306/54143209-ab4fea80-4429-11e9-8d26-e90bc61e8d02.png)

ildyria:
Note, this still means that this info should not be returned by the server when in public mode.

PR #158, opened 2019-10-17
Skip empty exif data fields

tmp-hallenser:
Restrict the shown exif data rows to:
* exif attribute has a value OR
* exif attribute is editable

If an attribute is empty and not editable -> skip it.

For video files, make and type are added to the list of available exif attributes.


ildyria:
@kamil4 also related: https://github.com/LycheeOrg/Lychee-front/issues/49 :)

ildyria:
I am merging this as it is a sane default behavior. :)

PR #160, opened 2019-10-17
Search by clicking on tags and modify search to change URL

tmp-hallenser:
**1. Search**
* Change URL to format '#search/searchstring' for searches
* set title for result page to 'Search results'

**2. Search by clicking on tags**
* Click on a photo tag triggers the search for this tag (only works if search is available, e.g. not for public if public_search = false)
* tags are always visible, but only editable when photo is editable

### DIFF
--- a/scripts/main/lychee.js
+++ b/scripts/main/lychee.js
@@ -366,6 +366,10 @@ lychee.load = function(autoplay = true) {
 			// Search has been triggered
 			search_string = decodeURIComponent(photoID);
 
+			if(search_string.trim()!=="") {
+				// do nothing on "only space" search strings
+				return;
+			}
 			// If public search is diabled -> do nothing
 			if (lychee.publicMode===true && !lychee.public_search) {
 				loadingBar.show('error', lychee.locale['ERROR_SEARCH_DEACTIVATED']);
@@ -414,6 +418,8 @@ lychee.load = function(autoplay = true) {
 			mapview.open();
 			lychee.footer_hide();
 
+		} else if(albumID=='search') {
+			// search string is empty -> do nothing
 		} else {
 
 			// Trash data
--- a/scripts/main/lychee.js
+++ b/scripts/main/lychee.js
@@ -372,11 +372,6 @@ lychee.load = function(autoplay = true) {
 				return;
 			}
 
-			if (visible.photo()) view.photo.hide();
-			if (visible.sidebar()) sidebar.toggle();
-
-			header.setMode('albums');
-
 			header.dom('.header__search').val(search_string);
 			search.find(search_string);
 
--- a/scripts/main/lychee.js
+++ b/scripts/main/lychee.js
@@ -303,7 +303,7 @@ lychee.logout = function() {
 lychee.goto = function(url = '', autoplay = true) {
 
 	if(url===false) url = '';
-	
+
 	url = '#' + url;
 
 	history.pushState(null, null, url);
@@ -360,6 +360,27 @@ lychee.load = function(autoplay = true) {
 			mapview.open(albumID);
 			lychee.footer_hide();
 
+
+		} else if (albumID=='search') {
+
+			// Search has been triggered
+			search_string = decodeURIComponent(photoID);
+
+			// If public search is diabled -> do nothing
+			if (lychee.publicMode===true && !lychee.public_search) {
+				loadingBar.show('error', lychee.locale['ERROR_SEARCH_DEACTIVATED']);
+				return;
+			}
+
+			if (visible.photo()) view.photo.hide();
+			if (visible.sidebar()) sidebar.toggle();
+
+			header.setMode('albums');
+
+			header.dom('.header__search').val(search_string);
+			search.find(search_string);
+
+			lychee.footer_show();
 
 		} else {
 			// Show photo
--- a/scripts/main/lychee.js
+++ b/scripts/main/lychee.js
@@ -328,7 +328,6 @@ lychee.load = function(autoplay = true) {
 	let photoID	= '';
 	let hash	= document.location.hash.replace('#', '').split('/');
 
-	$('.no_content').remove();
 	contextMenu.close();
 	multiselect.close();
 
@@ -344,6 +343,7 @@ lychee.load = function(autoplay = true) {
 				loadingBar.show('error', lychee.locale['ERROR_MAP_DEACTIVATED']);
 				return;
 			}
+			$('.no_content').remove();
 			// show map
 			// albumID has been stored in photoID due to URL format #map/albumID
 			albumID = photoID;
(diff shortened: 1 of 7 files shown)


scripts/main/lychee_locale.js
```diff
@@ -238,6 +239,7 @@ lychee.locale = {
 		'ERROR_UNKNOWN'				: 'Something unexpected happened. Please try again and check your installation and server. Take a look at the readme for more information.',
 		'ERROR_LOGIN'				: 'Unable to save login. Please try again with another username and password!',
 		'ERROR_MAP_DEACTIVATED'				: 'Map functionality has been deactivated under settings.',
+		'ERROR_SEARCH_DEACTIVATED'				: 'Search functionality has been deactivated under settings.',
```
kamil4: I'm generally somewhat skeptical about adding elaborate notifications for cases that simply aren't supposed to happen. I'm just not sure if the amount of effort required to do it properly (adding them to all the language files on the server side) is worth it.

But it's your time and you choose how to use it :wink:.

scripts/main/sidebar.js
```diff
@@ -205,22 +224,17 @@ sidebar.createStructure.photo = function(data) {
...
-			type  : sidebar.types.TAGS,
-			value : build.tags(data.tags),
-			editable
-		}
-
-	} else {
-
-		structure.tags = {}
+	// Always create tags section - behaviour for editing
+	//tags handled when contructing the html code for tags
```
kamil4: This begs the question -- are tags supposed to be private or public?

I have no opinion on this so I'm fine with your change but I thought we should check with others...

scripts/main/lychee.js
```diff
@@ -358,6 +358,22 @@ lychee.load = function(autoplay = true) {
 			lychee.footer_hide();
 
 
+		} else if (albumID=='search') {
```
kamil4: This does not work correctly for me when the search bar at the top is empty and I'm activating it to type a search query. With your change, I'm being immediately redirected to an empty album, probably because `photoID` is empty?

scripts/main/sidebar.js
```diff
@@ -73,6 +80,18 @@ sidebar.bind = function() {
...
+sidebar.triggerSearch = function(search_string) {
+
+	// If public search is diabled -> do nothing
+	if (lychee.publicMode===true && !lychee.public_search) {
+		// Do not display an error -> just do nothing to not confuse the user
+		return;
+	}
+
+	// We're either logged in or public search is allowed
+	lychee.goto('search/' + encodeURIComponent(search_string));
```
kamil4: I think you need to reset `search.hash` to `null` before this call (the way `search.reset` does it).

I'm seeing a problem where clicking on a tag only works the first time. If I click on the same tag again, without any other searches in-between, nothing happens, and the reason is that `search.hash` was not reset in the meantime.

scripts/main/sidebar.js
d7415: I don't have time for a proper review, but I'd say that if you can see the photo there should be no harm in seeing the tags. For the reverse, I'd say that you should only be able to see tags if you have access to at least one photo with that tag.

ildyria: Agreed with @d7415

scripts/main/lychee_locale.js
```diff
@@ -31,6 +31,7 @@ lychee.locale = {
 		'SHARED_ALBUMS'		: 'Shared albums',
 		'ALBUMS'			: 'Albums',
 		'PHOTOS'			: 'Pictures',
+		'SEARCH_RESULTS'			: 'Search results',
```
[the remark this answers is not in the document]
tmp-hallenser: This has already been done and merged ( https://github.com/LycheeOrg/Lychee-Laravel/pull/374 )

scripts/main/lychee.js
tmp-hallenser: should be fixed

scripts/main/sidebar.js
tmp-hallenser: should be resolved

scripts/main/lychee_locale.js
tmp-hallenser: > For the reverse, I'd say that you should only be able to see tags if you have access to at least one photo with that tag.

@d7415 Could you explain in more detail what you mean? 
Let's assume we've got photo 1 (tag A), photo 2 (tag A & B) and photo 3 (no tag). All are pubic.
Do you mean:
(1) Tag A is only displayed at photos to user (in public mode), if at least one of the photos visible to him has tag A? (This would be imho trivial since you can't view tags without viewing a photo <- correct me if I'm wrong). 
(2) The section Tags is only shown, if at least one photo visible to the user (in public mode) has a tag -> This would require an additional db query for every photo and backend changes -> I'd vote against this because of (a) simplicity (what's the harm of showing an empty tag section) and (b) performance (I think I read somewhere that there has been recently an attempt to reduce the number of db queries to increase performance) 
(3) something completely different

scripts/main/lychee_locale.js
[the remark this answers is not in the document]
tmp-hallenser: I've already spent the time and it has been merged :)

scripts/main/sidebar.js
tmp-hallenser: > For the reverse, I'd say that you should only be able to see tags if you have access to at least one photo with that tag.

@d7415 Could you explain in more detail what you mean?
Let's assume we've got photo 1 (tag A), photo 2 (tag A & B) and photo 3 (no tag). All are pubic.
Do you mean:
(1) Tag A is only displayed at photos to user (in public mode), if at least one of the photos visible to him has tag A? (This would be imho trivial since you can't view tags without viewing a photo <- correct me if I'm wrong).
(2) The section Tags is only shown, if at least one photo visible to the user (in public mode) has a tag -> This would require an additional db query for every photo and backend changes -> I'd vote against this because of (a) simplicity (what's the harm of showing an empty tag section) and (b) performance (I think I read somewhere that there has been recently an attempt to reduce the number of db queries to increase performance)
(3) something completely different

scripts/main/lychee_locale.js
tmp-hallenser: I've already spent the time and it has been merged :)

scripts/main/sidebar.js
d7415: I meant 1. This was in a broader sense, in case we list them elsewhere in the future (and in case you'd implemented that here - I still haven't had a change to review it!)

To confirm:

- no concerns with viewing a photo's tags from photo view.
-If a list of tags is going to be used elsewhere (e.g. autocomplete when adding tags) it should only show a tag if the user has access to a photo that uses that tag.

tmp-hallenser: Thanks for the explanation. This make sense and is no problem here.

kamil4:
I pushed some more tweaks to make editing the search string more seamless. This should now be ready for review!

ildyria:
Rebased on Master for CI check.

PR #163, opened 2019-10-30
Support drag-and-drop or paste in well-defined UI state only

kamil4:
This is an overdue patch to clean up when we allow drag-and-drop or image paste operations. In my opinion, they should be supported only in albums or album views, when there are no dialog windows/menus/etc. being displayed. Otherwise there's a good chance that the operation was performed accidentally.

### DIFF
--- a/scripts/main/init.js
+++ b/scripts/main/init.js
@@ -136,13 +136,9 @@ $(document).ready(function() {
 	.on('dragover', function() { return false }, false)
 	.on('drop', function(e) {
 
-		if (!album.isUploadable()) {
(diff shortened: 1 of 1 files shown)


ildyria:
Did not test, but the change sound rather sane if I step through it in my mind.

kamil4:
My thinking was basically that upload via drag-and-drop or paste should be allowed only in situations when you can access the "+" menu to do it manually. The code was doing complicated things such as closing the open dialog windows and such. Better to simply ignore the event in such cases so that the user can finish what they were doing and then they can retry the upload if that's what they actually intended.

PR #164, opened 2019-10-30
Correct the updates of the counts in the sidebar

kamil4:
Another overdue patch. Cleans up and fixes missing cases in the updating of subalbum/image/video counts in the sidebar after delete operations. In the process gets rid of a completely unnecessary `album.json.num` so that in the future we can remove it from the server as well (it's only needed for smart albums).


ildyria:
Rebased on master for CI Checks.

PR #168, opened 2019-11-14
Frame refresh

kamil4:
While fixing https://github.com/LycheeOrg/Lychee-Laravel/issues/390 I noticed that the refresh of the image would often flicker, with new image being rendered too late. I'm not sure if I fixed it 100% but I sure made it better...

I also cleaned up the code a bit in the process, including fixing the case when full-size images are disabled.


ildyria:
Tested.

PR #178, opened 2020-01-16
Bugfix

### LINKED ISSUE #177: Collapsing sidebar style issues
### Detailed description of the problem

When closing the left sidebar the label text drops to a new line below the icon because of the collapsing div. It seems applying a `min-width: 210px;` to the class `.linkMenu` fixes this issue.

Another thing I noticed is that all but the Settings menu items have a preceeding whitespace which causes the items to look misaligned. 

Apologies for being nitpicky :) 

Open and close the sidebar. You can see a glimpse of the text being pushed to a newline before the sidebar completely closes.

### Browser and system

Chromium Version 79.0.3945.117 (Official Build) Arch Linux (64-bit)
Firefox Version 72.0.1 Arch Linux (64-bit)

bluser86:
Fix #177

Added a min-width property for blocklevel left menu items so that collapsing
the menu doesn't cause the menu item labels to cascade to a new line.

Also removed spaces between menu items and icons because this causes
misalignment.

### DIFF
--- a/scripts/main/leftMenu.js
+++ b/scripts/main/leftMenu.js
@@ -23,18 +23,18 @@ leftMenu.build = function () {
 	if (lychee.api_V2)
 	{
 		html += lychee.html`
-		<a class="linkMenu" id="button_users">${    build.iconic('person') } ${             lychee.locale['USERS'] } </a>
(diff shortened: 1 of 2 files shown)


bluser86:
I actually changed my mind regarding the min-width as this might have implications when the menu needs to be resized for responsiveness. Instead I opted for using a `white-space: nowrap;` setting on the link menu to prevent wrapping that way.

Issue #174, opened 2020-01-12
[Mobile] Toggling of header


tmp-hallenser:
### Detailed description of the problem
Toggling header by tapping on photo does not work 

### Steps to reproduce the issue
1. Open a photo on a mobile device (Android, iOS)
2. Wait for header to hide 
3. Tap on photo -> header appears again
4. Tap again on photo -> header should disappear, but doesn't

### Analysis
It did some debugging, but was not able to resolve the issue.

The following code line responsible for step 4:
https://github.com/LycheeOrg/Lychee-front/blob/9c2c5bf07b97afa179248e9bffd5d55d77d29403/scripts/main/init.js#L106

Function `header.hide(..)`gets called and should hide the header. The responsible lines of code are:
https://github.com/LycheeOrg/Lychee-front/blob/9c2c5bf07b97afa179248e9bffd5d55d77d29403/scripts/main/header.js#L131-L132

Both lines are called, but have no effect.

PR #188, opened 2020-04-11
Fix toggle of header

tmp-hallenser:
Fix toggling of header in 2 cases

1. single touch does not toggle header (#174)
Event 'mousemove' gets called after a single touch and 'mousemove' makes header appear; Solution is to prevent event 'mousemove' to be called after 'touchend'

2. swipe forward / backward toogles header
Swiping forward / backward triggers 'touchend' event, which toggles the header. Solution is to prevent toggling if 'swipeend' triggered loading of previous/next photo

### DIFF
--- a/scripts/main/init.js
+++ b/scripts/main/init.js
@@ -105,9 +105,20 @@ $(document).ready(function() {
 
 			// Fullscreen on mobile
 			.on('touchend', '#imageview #image', function(e) {
-				if (swipe.obj==null || (swipe.offsetX>=-5&&swipe.offsetX<=5)) {
-					if (visible.header()) header.hide(e);
-					else                  header.show()
+
+				// prevent triggering event 'mousemove'
+				e.preventDefault();
+
+				if ((swipe.obj==null) || ((Math.abs(swipe.offsetX)<=5 && Math.abs(swipe.offsetY)<=5))) {
+					// Toogle header only if we're not moving to next/previous photo;
+					// In this case, swipe.preventNextHeaderToggle is set to true
+					if((swipe.preventNextHeaderToggle==null) || (swipe.preventNextHeaderToggle==false)) {
+						if (visible.header()) header.hide(e);
+						else                  header.show();
+					}
+
+					// For next 'touchend', behave again as normal and toogle header
+					swipe.preventNextHeaderToggle = false;
 				}
 			});
 		$('#imageview')
--- a/scripts/main/init.js
+++ b/scripts/main/init.js
@@ -109,12 +109,15 @@ $(document).ready(function() {
 				// prevent triggering event 'mousemove'
 				e.preventDefault();
 
-				if ((swipe.obj==null) || ((Math.abs(swipe.offsetX)<=5 && Math.abs(swipe.offsetY)<=5))) {
+				if ((typeof swipe.obj === undefined) || (Math.abs(swipe.offsetX)<=5 && Math.abs(swipe.offsetY)<=5)) {
 					// Toogle header only if we're not moving to next/previous photo;
 					// In this case, swipe.preventNextHeaderToggle is set to true
-					if((swipe.preventNextHeaderToggle==null) || (swipe.preventNextHeaderToggle==false)) {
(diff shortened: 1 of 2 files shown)


ildyria:
:astonished:  Congrat for finding that one.

tmp-hallenser:
Btw, I am across a couple of code snipped where I didn't see the logic:

1. What's the purpose of this line?
https://github.com/LycheeOrg/Lychee-front/blob/fe0cb5415748340fc6529f5c623bf289c1cdd33f/scripts/main/init.js#L102
The events in the next section will not be triggered if you don't have a touch device.

2. Duplicates
There seem to be a duplicates:
https://github.com/LycheeOrg/Lychee-front/blob/fe0cb5415748340fc6529f5c623bf289c1cdd33f/scripts/view/main.js#L191 
https://github.com/LycheeOrg/Lychee-front/blob/fe0cb5415748340fc6529f5c623bf289c1cdd33f/scripts/main/init.js#L28 
and
https://github.com/LycheeOrg/Lychee-front/blob/7b21ec0b5fc573ddf72f02f7f80a5a9aee51ea80/scripts/main/lychee.js#L659
https://github.com/LycheeOrg/Lychee-front/blob/7b21ec0b5fc573ddf72f02f7f80a5a9aee51ea80/scripts/frame/main.js#L60

3. Hybrid devices
We bind many event to either click or touchend by this function:
https://github.com/LycheeOrg/Lychee-front/blob/7b21ec0b5fc573ddf72f02f7f80a5a9aee51ea80/scripts/frame/main.js#L60
This falls short in case of convertible laptops, e.g. the X1 Carbon Yoga series. Why do we need this function at all? Shouldn't click be also triggered on a touch device (although a bit later in the event cascade). Furthermore, as a convertible user, I would expect the app to behave the same if I use the mouse of my finder to navigate (in particular these lines https://github.com/LycheeOrg/Lychee-front/blob/fe0cb5415748340fc6529f5c623bf289c1cdd33f/scripts/main/init.js#L104-L117)

ildyria:
Tested. Did not see any problem.

ildyria:
> Btw, I am across a couple of code snipped where I didn't see the logic:
> 
> 1. What's the purpose of this line?
>    https://github.com/LycheeOrg/Lychee-front/blob/fe0cb5415748340fc6529f5c623bf289c1cdd33f/scripts/main/init.js#L102
>    
>    The events in the next section will not be triggered if you don't have a touch device.

No clue.

> 2. Duplicates
>    There seem to be a duplicates:
>    https://github.com/LycheeOrg/Lychee-front/blob/fe0cb5415748340fc6529f5c623bf289c1cdd33f/scripts/view/main.js#L191
>    
>    https://github.com/LycheeOrg/Lychee-front/blob/fe0cb5415748340fc6529f5c623bf289c1cdd33f/scripts/main/init.js#L28
>    
>    and
>    https://github.com/LycheeOrg/Lychee-front/blob/7b21ec0b5fc573ddf72f02f7f80a5a9aee51ea80/scripts/main/lychee.js#L659
>    
>    https://github.com/LycheeOrg/Lychee-front/blob/7b21ec0b5fc573ddf72f02f7f80a5a9aee51ea80/scripts/frame/main.js#L60

Gotta remove them then.

> 3. Hybrid devices
>    We bind many event to either click or touchend by this function:
>    https://github.com/LycheeOrg/Lychee-front/blob/7b21ec0b5fc573ddf72f02f7f80a5a9aee51ea80/scripts/frame/main.js#L60
>    
>    This falls short in case of convertible laptops, e.g. the X1 Carbon Yoga series. Why do we need this function at all? Shouldn't click be also triggered on a touch device (although a bit later in the event cascade). Furthermore, as a convertible user, I would expect the app to behave the same if I use the mouse of my finder to navigate (in particular these lines https://github.com/LycheeOrg/Lychee-front/blob/fe0cb5415748340fc6529f5c623bf289c1cdd33f/scripts/main/init.js#L104-L117
>    )

Feel free to refactor that, I don't have a strong opinion. :smiley:

PR #195, opened 2020-04-25
Put travis node version back to automatic

d7415:
Now that the dependencies should be fixed.


d7415:
This is worth remembering for node 15 etc, but it's probably better for travis to catch an incompatibility rather than wait for us to notice the upgrade, update `.travis.yml` and *then* find it's broken.

PR #197, opened 2020-05-03
Feature/orientation

alex-phillips:
Goes with matching PR in the Lychee-front repo.

If the rotation metadata is set on a photo (meaning that we failed to auto-rotate the image server-side), then the frontend can account for rotating the image as necessary in the UI.

### DIFF
--- a/scripts/main/build.js
+++ b/scripts/main/build.js
@@ -134,6 +134,7 @@ build.photo = function (data, disabled = false) {
 	let isVideo = data.type && data.type.indexOf('video') > -1;
 	let isRaw = data.type && data.type.indexOf('raw') > -1;
 	let isLivePhoto = (data.livePhotoUrl!=='' && data.livePhotoUrl!==null);
+	let orientationClass = `orientation-${data.orientation}`;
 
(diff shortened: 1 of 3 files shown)


alex-phillips:
I need to wrap up a few additional commits which handle the orientation of single image view before this can be accepted.

alex-phillips:
@ildyria I may need some suggestions on how to handle the imagerotation with 'imageview'. The rotation doesn't swatch the width and height dimensions so it doesn't end up fitting. 

Another option which would solve most issues is if we can auto-rotate the images when we create the smaller sizes.

ildyria:
> Another option which would solve most issues is if we can auto-rotate the images when we create the smaller sizes.

I would rather go that way to be honest. This would lower the load of changes etc in the front end.

d7415:
> > Another option which would solve most issues is if we can auto-rotate the images when we create the smaller sizes.
> 
> I would rather go that way to be honest. This would lower the load of changes etc in the front end.

Agreed. It also means the load is one-off per image rather than every image load and should reduce code complexity.

PR #204, opened 2020-06-28
hide live photo download

ikosa:
to hide live photo on download modal when not present


ildyria:
Code wise it looks sane, but I don't have pictures to test it with.

ikosa:
There are sample photos here:https://github.com/LycheeOrg/Lychee/issues/378 i
test with them. And you can test with any photo to see that download modal
has "Video part of live-photo" option in every photo that are not
live-photo.


Benoît Viguier <notifications@github.com>, 3 Tem 2020 Cum, 17:06 tarihinde
şunu yazdı:

PR #205, opened 2020-07-09
Support fire tv

tmp-hallenser:
This PR adds basic support to view Lychee on a FireTV as a WebApp.

In a nutshell:
* Navigation is done using the 'tabindex' property of elements
* We need to enable/disable this features as element can be focused even if they are not visible 
* Several navigation elements are disabled since they are not needed on a TV
* Device specific CSS (using white to highlight nav elements) used
* FireTV doesn't like transparency and several layers of elements on top of each other (performance) 
* UI for Desktop and mobile remains unchanged

### DIFF
--- a/scripts/main/header.js
+++ b/scripts/main/header.js
@@ -24,7 +24,7 @@ header.bind = function() {
 
 		if ($(this).hasClass('header__title--editable')===false) return false;
 
-		if(lychee.enable_contextmenu_header==false) return false;
+		if(lychee.enable_contextmenu_header===false) return false;
 
 		if (visible.photo()) contextMenu.photoTitle(album.getID(), photo.getID(), e);
 		else                 contextMenu.albumTitle(album.getID(), e)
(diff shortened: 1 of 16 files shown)


scripts/main/header.js
```diff
@@ -250,41 +347,100 @@ header.setMode = function(mode) {
...
-				$('#button_more').hide();
+				e = $('#button_more');
+				e.hide();
+				tabindex.makeUnfocusable(e);
 			}
 
+			// Remove buttons if needed
+			if(!lychee.enable_button_visibility) {
+				e = $('#button_visibility', '.header__toolbar--photo');
+				e.hide();
```
ildyria, line 398: Can't you use https://api.jquery.com/remove/ instead of `.hide()` ?

I feel this would simplify the dom. :)

Vorlif:
Hey, can you briefly explain how I can link/open lychee as a web app under fire tv?

d7415:
> Hey, can you briefly explain how I can link/open lychee as a web app under fire tv?

It sounds like @tmp-hallenser might have some experience, but at a glance, https://developer.amazon.com/docs/fire-tv/getting-started-with-web-apps.html

PR #210, opened 2020-07-27
Don't rely on cameraDate

kamil4:
This fixes a regression I've noticed where we no longer print the take date of photos underneath their names in the album view. This is because the server no longer sends the `cameraDate` field. We can rely on the `takedate` instead though.


ildyria:
> This is because the server no longer sends the `cameraDate` field. We can rely on the `takedate` instead though.

Yes, I noticed server side that we were just setting `cameraDate` if `takedate` was not `''`, So I removed it when I changed the backend for an easier addition of smart albums.

I though I fixed most of the places in the front end.

PR #216, opened 2020-08-14
fix: 422 error when create new album in SmartAlbum

YoshinoriN:
## Description

Fix: https://github.com/LycheeOrg/Lychee/issues/658

## Changes

Set parent album to `0` when creating a new album in SmartAlbum.
I referred to the below comment.

>  https://github.com/LycheeOrg/Lychee/issues/658#issuecomment-657112143

---

Thank you :)

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -256,7 +256,7 @@ album.add = function (IDs = null, callback = null) {
 			parent_id: 0
 		};
 
-		if (visible.albums()) {
+		if (visible.albums() || album.isSmartID(album.json.id)) {
 			params.parent_id = 0;
 		} else if (visible.album()) {
 			params.parent_id = album.json.id;


kamil4:
Wouldn't it make more sense to disable subalbum creation menu item in the front end when in a smart album?

YoshinoriN:
> Wouldn't it make more sense to disable subalbum creation menu item in the front end when in a smart album?

I think so too.
Should I fix it like that?

kamil4:
I'm curious what @ildyria and @d7415 think...

In my opinion smart albums should be effectively read-only, with the possible exception of photo uploads to the Unsorted one. The bug being fixed here is just one example of the problems the current behavior causes, but I'm sure there are more. E.g., I just played with it for 30 seconds and I found that clicking the Delete button gives "Album json not found" error.

d7415:
> In my opinion smart albums should be effectively read-only, with the possible exception of photo uploads to the Unsorted one. 

I'm inclined to agree, including that exception. It reduces the number of edge cases and special handling we need, which is good!

ildyria:
> > In my opinion smart albums should be effectively read-only, with the possible exception of photo uploads to the Unsorted one.
> 
> I'm inclined to agree, including that exception. It reduces the number of edge cases and special handling we need, which is good!

this.

kamil4:
I'll merge this since it fixes a server error notification but as we discussed a more comprehensive change should be implemented in the future.

YoshinoriN:
> I'll merge this since it fixes a server error notification but as we discussed a more comprehensive change should be implemented in the future.

I'll try to implement it this weekend :)

kamil4:
I'm guessing that modifying `album.isUploadable` will get you 90% there. Right now it returns `true` for smart albums, I believe.

PR #219, opened 2020-08-27
Feature/#48 album by tags

Aleshus:
Added support for tag albums


ildyria (approved):
LGTM

scripts/main/album.js
```diff
@@ -43,6 +43,10 @@ album.getID = function () {
 
 };
 
+album.isTagAlbum = function() {
+	return album.json['tag_album'] === '1';
```
kamil4, line 47: Is there a reason you are using array syntax rather than `album.json.tag_album`? This is not a trick question; I'm genuinely curious :smiley:.

scripts/main/album.js
```diff
@@ -298,6 +302,100 @@ album.add = function (IDs = null, callback = null) {
...
+		basicModal.close();
+
+		let params = {
+			title: data.title,
+			tags: data.tags,
+			parent_id: 0
+		};
+
+		api.post('Album::addByTags', params, function (_data) {
+			const isNumber = (n) => (!isNaN(parseInt(n, 10)) && isFinite(n));
```
kamil4, line 318: I'm sure @ildyria loves this syntax. He's been itching to use something like this on the server side :wink:.

scripts/main/album.js
```diff
@@ -298,6 +302,100 @@ album.add = function (IDs = null, callback = null) {
...
+			if (_data !== false && isNumber(_data)) {
+				if (IDs != null && callback != null) {
+					callback(IDs, _data, false); // we do not confirm
```
kamil4, line 321: What is that callback for? I don't think you use it?

scripts/main/header.js
```diff
@@ -279,11 +279,21 @@ header.setMode = function(mode) {
 				tabindex.makeUnfocusable(e);
 			}
 
-			if (albumID==='s' || albumID==='f' || albumID==='r') {
-				$('#button_info_album, #button_trash_album, #button_visibility_album, #button_move_album').hide();
-				$('.button_add, .header__divider', '.header__toolbar--album').show();
-				tabindex.makeFocusable($('.button_add, .header__divider', '.header__toolbar--album'));
-				tabindex.makeUnfocusable($('#button_info_album, #button_trash_album, #button_visibility_album, #button_move_album'));
+			if (albumID==='starred' || albumID==='unsorted' || albumID === 'public' || albumID==='recent' || album.isTagAlbum()) {
```
kamil4, line 282: I'm guessing that in addition to adding support for tag albums you were fixing "regular" smart album support here? I'm confused though because `#button_visibility_album` is now unconditionally shown (in the next line) whereas before it was hidden. Also, you included `unsorted` in this `if` whereas before it was handled in the next `else if (albumID==='0')`, which you've kept (see below), but which will now be unreachable. So something's definitely not right here...

I'm a little surprised by the overall approach here. Why didn't you simply replace the broken `'s'`, `'f'`, and `'r'` here (and `'0'` below) with `'starred'`, `'public'`, `'recent'`, and `'unsorted'`, keeping the insides of those `if`s as-is, and add a new condition just for tag albums further down?

scripts/main/album.js
```diff
@@ -298,6 +302,100 @@ album.add = function (IDs = null, callback = null) {
...
+album.addByTags = function (IDs = null, callback = null) {
+
+	const action = function (data) {
+
+		basicModal.close();
+
+		let params = {
+			title: data.title,
+			tags: data.tags,
+			parent_id: 0
```
kamil4, line 314: I don't think you need that `parent_id`?

scripts/main/album.js
Aleshus: there is no reason. Dot variant is more pretty, you are right.

scripts/main/album.js
Aleshus: removed callback and IDs parameter too

scripts/main/header.js
Aleshus: Ok I did not understand that albumID=='unsorted' is same as albumID=='0' and I thought that unsorted album should have same funcions as smart albums (but it is not true). So I agree that unsorted shouldn't be in the first condition and I already removed it and changed that zero to unsorted.

Second point is true too. At first I thought that tag albums should haś same behaviour as other smart albums. But it is not true, for example in trash button it is different. So your idea is better, I will refactor it.. thanks for it!

Aleshus: But is it ok, that on smart album is visible add (+) button ?

scripts/main/album.js
kamil4: I fixed this one...

scripts/main/header.js
kamil4: I had to fix the publicMode case (we were still displaying the visibility and trash icons).

Answering your question, that's how it's always been. I think we should restrict it (photo uploads to smart albums make some limited sense; subalbum creation definitely not). But let's deal with that in a separate issue; otherwise we'll never get yours out the door! :smile:

kamil4:
This should be good to go now

kamil4 (approved):

PR #220, opened 2020-09-01
some dependencies move from dependencies to devDependencies

YoshinoriN:
## Description

Move some dependencies from `dependencies` to `devDependencies` in `package.json`

## Reason

The `devDependencies` are just for development dependency that like gulp (task runner), babel (transpiler), etc...
It is similar to `require` and `require-dev` relationship in `composer.json` for PHP.

IMHO, it is better to separate them for developers.

Thank you :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,17 +17,7 @@
     "node": ">= 10"
   },
   "dependencies": {
-    "babel-preset-env": "^1.7.0",
     "basicmodal": "github:LycheeOrg/basicmodal#master",
-    "gulp": "^4.0.0",
-    "gulp-autoprefixer": "^6.0.0",
-    "gulp-babel": "^6.1.2",
-    "gulp-clean-css": "^3.10.0",
-    "gulp-concat": "^2.6.1",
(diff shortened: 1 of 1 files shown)


ildyria:
To be honest, I'm not sure it matters as we are just using node to compile our JS. :sweat_smile:

d7415:
> To be honest, I'm not sure it matters as we are just using node to compile our JS. sweat_smile

That was my immediate reaction. If you're using this repo and aren't developing, you're doing it wrong!

YoshinoriN:
IMHO, if separate `dependencies` and `devDependencies` we can understand a little easier which packages affect application behavior. I think this is also useful when upgrading a package version.

Generally, We have to check & test application behavior when update the package's version. But, `devDependencies` are no need. Because `devDependencies` are just for development, not change the application behavior.

But, I cant come up with other advantages.

ildyria:
> (aside from apple-touch-icon-* which should be in dist...)

Actually they are in `img/` and it is fixed with https://github.com/LycheeOrg/Lychee/pull/721

PR #221, opened 2020-09-02
Make photo height configurable via CSS

kamil4:
Doesn't change the default, simply makes it configurable via user.css.

### DIFF
--- a/styles/main/_justified_layout.scss
+++ b/styles/main/_justified_layout.scss
@@ -13,7 +13,7 @@
 
 .justified-layout > .photo {
 	position: absolute;
-	height: 320px;
+	--lychee-default-height: 320px;
 	margin: 0;
 }
 
--- a/styles/main/_justified_layout.scss
+++ b/styles/main/_justified_layout.scss
@@ -13,6 +13,7 @@
 
 .justified-layout > .photo {
 	position: absolute;
+	height: 320px;
 	margin: 0;
 }
 
--- a/scripts/main/view.js
+++ b/scripts/main/view.js
@@ -372,7 +372,7 @@ view.album = {
 				let layoutGeometry = require('justified-layout')(ratio, {
 					containerWidth: containerWidth,
 					containerPadding: 0,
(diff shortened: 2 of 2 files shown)


styles/main/_justified_layout.scss
```diff
@@ -13,6 +13,7 @@
 
 .justified-layout > .photo {
 	position: absolute;
+	height: 320px;
```
d7415, line 16: If our default `small_max_height` is 360, should this not be 360 too? Am I missing something?

kamil4: Well, the default of the justified-layout package that we use is 320px, and that's what we've been using in Lychee all along. I didn't want to change it since it would affect the look of people's existing albums.

As far as I can tell, the default value of `small_max_height` is arbitrary. It does not influence the layout, it simply determines the rendering quality. Perhaps it was set to 360 by mistake, or somebody decided that that would be "good enough" (you may know better than I do; that value predates my involvement in the project).

I agree that setting it to 320 would be less arbitrary but it would not necessarily be optimal. To keep the right margin justified, justified-layout can stretch an image by up to 2x, so to ensure the optimal rendering quality we should set it to 640. In a way, we already generate those higher-resolution variants and store in "@2x", and even on low-DPI displays the browser will currently use them for images stretched beyond 360px. But on HiDPI displays we do need to set `small_max_height` higher than 320. In principle, no value is guaranteed to be optimal as displays can be arbitrarily dense of course; we currently make provisions for double density and that's it.

kamil4:
I pushed a commit to use a custom CSS property name rather than the default `height` since we actually change `height` while laying out so if we needed to rejustify (after a window size change) weird things would happen.

styles/main/_justified_layout.scss
d7415: > the default of the justified-layout package that we use is 320px

Good enough for me :)

I'm pretty sure all those numbers were chosen arbitrarily. And that makes sense re 32/360/whatever. We have to draw the line somewhere.

ildyria (approved):

PR #223, opened 2020-09-15
Match the title limits of the server

kamil4:
Fixes https://github.com/LycheeOrg/Lychee/issues/723


d7415 (approved):
Looks simple enough. I haven't built this, but I did manually change the rename field to check that the server was happy with 100 characters.

kamil4:
Yeah, I even tried with some Unicode characters to make sure that the limit wasn't 100 _bytes_.

ildyria (approved):

PR #232, opened 2020-12-17
fix of album undefined

nickelsound:


ildyria:
@nickelsound please check, if it works on your side I will merge it.

kamil4:
Sorry to butt in. I wasn't around when this PR was first merged so I couldn't ask then.

I think it's a very nice feature but I've been wondering why it uses `localStorage`? Couldn't it simply store the positions in an internal JS variable (maybe even in the `albums` tree)? What's the value of storing the album positions, effectively, permanently? Or am I misunderstanding how this works?

nickelsound:
It should survive a browser refresh, or that was at least the purpose. The storage removes the index automatically if the position in the album is top. So it cleans itself.

nickelsound:
@ildyria  checked, it works all right

PR #233, opened 2020-12-18
Add Sensitive folders.

ildyria:
related https://github.com/LycheeOrg/Lychee/pull/822


README.md
```diff
@@ -61,6 +61,49 @@ npm install
...
+    echo "${YELLOW}Please install prettier, e.g.:${NO_COLOR}"
+    echo ""
+    echo "  npm install"
+    echo ""
+fi
+
+echo "\n${GREEN}pre commit hook finish${NO_COLOR}\n"
+```
+
+This can easilly be installed by doing:
```
d7415, line 100:
```suggestion
This can easily be installed by doing:
```

scripts/main/album.js
```diff
@@ -775,6 +783,12 @@ album.setPublic = function (albumID, e) {
 			}
 		});
 
+		if (album.json.nsfw === "1") {
+			$('.basicModal .switch input[name="nsfw"]').prop("checked", true);
```
d7415, line 787: I'm guessing we can't just
```js
$('.basicModal .switch input[name="nsfw"]').prop("checked", album.json.nsfw === "1");
```
?

scripts/main/album.js
```diff
@@ -852,6 +890,22 @@ album.setPublic = function (albumID, e) {
 	});
 };
 
+album.setNSFW = function (albumID, e) {
```
d7415, line 893: Would `toggleNSFW` fit better than `setNSFW` here?

scripts/main/lychee.js
```diff
@@ -268,6 +284,13 @@ lychee.init = function () {
 			lychee.swipe_tolerance_x = validatedSwipeToleranceX;
 			lychee.swipe_tolerance_y = validatedSwipeToleranceY;
 
+			lychee.nsfw_visible = (data.config.nsfw_visible && data.config.nsfw_visible === "1") || false;
```
d7415, line 287: Not for this PR, but in general there is a *lot* of duplication here :(

scripts/main/lychee_locale.js
```diff
@@ -141,6 +141,10 @@ lychee.locale = {
 	ALBUM_FULL_EXPL: "Full size pictures are available",
 	ALBUM_HIDDEN: "Hidden",
 	ALBUM_HIDDEN_EXPL: "Only people with the direct link can view this album.",
+	ALBUM_MARK_NSFW: "Mark album as sensitive",
+	ALBUM_UNMARK_NSFW: "Unmark album as sensitive",
+	ALBUM_NSFW: "Sensitive",
+	ALBUM_NSFW_EXPL: "Album is marked to contain sensitive content.",
```
d7415, line 147:
```suggestion
	ALBUM_NSFW_EXPL: "Album may contain sensitive content.",
```
or
```suggestion
	ALBUM_NSFW_EXPL: "Album contains sensitive content.",
```
?

scripts/main/settings.js
```diff
@@ -482,6 +482,29 @@ settings.changeLocationDecoding = function () {
 	});
 };
 
+settings.changeNSFWVisible = function () {
+	var params = {};
+	if ($("#NSFWVisible:checked").length === 1) {
```
d7415, line 487: That's ugly, but I don't have a better option :P

scripts/main/view.js
```diff
@@ -480,6 +496,16 @@ view.album = {
 		else sidebar.changeAttr("hidden", lychee.locale["ALBUM_SHR_YES"]);
 	},
 
+	nsfw: function () {
+		if (album.json.nsfw === "1") {
+			// Sensitive
+			$("#button_nsfw_album").addClass("active").attr("title", lychee.locale["ALBUM_UNMARK_NSFW"]);
+		} else {
+			// Unensitive
```
d7415, line 504:
```suggestion
			// Not Sensitive
```

d7415 (changes requested):
Untested. I can't see anything major. Mostly comments and strings.

scripts/main/album.js
ildyria: I didn't think about it, just copy pasted the pattern around.

scripts/main/album.js
ildyria: I would argue that toggle is more of an on-off temporary (a bit like the `h` short cut) while set is more in the length. Also that Set is more in line with the name of all the previously defined functions. :)

scripts/main/lychee.js
ildyria: I know :(

scripts/main/album.js
d7415: In my mind "set" is more saving a specific value, whereas this is changing it, *then* saving it.

scripts/main/album.js
d7415: I think it looks cleaner, but I don't really care.

scripts/frame/main.js
```diff
@@ -91,10 +91,10 @@ frame.refreshPicture = function () {
 		let srcset = "";
 		let src = "";
 		this.frame.photo = null;
-		if (data.medium != "") {
+		if (data.medium !== "") {
 			src = data.medium;
 
-			if (data.medium2x && data.medium2 != "") {
+			if (data.medium2x && data.medium2 !== "") {
```
kamil4, line 97: I realize you "inherited" this typo but surely it should say `medium2x` in the second condition as well?

scripts/main/lychee.js
```diff
@@ -42,6 +42,12 @@ let lychee = {
 	landing_page_enabled: false, // is landing page enabled ?
 	delete_imported: false,
 
+	nsfw_visible: true,
+	nsfw_visible_saved: true,
+	nsfw_blur: false,
+	nsfw_warning: false,
+	nsfw_warning_text: "<b>Sensitive content</b><br><p>This album contains sensitive content which some people may find offensive or disturbing.</p>",
```
kamil4, line 49: Shouldn't this be localized?

(as well as in two places further down in this file)

kamil4 (changes requested):
Looks good overall, apart from a couple minor points.

I still need to review the server changes and I'd also like to try it out (maybe tonight).

scripts/frame/main.js
ildyria: yes.

scripts/main/lychee.js
ildyria: Actually, no. It is a config text, so that the user can change it to its liking. :)
Though it is actually unnecessary as this is included in the blade template.

kamil4:
OK, so I tested this feature and there are a couple of things that bother me:
* Toggling the NSFW state using the warning triangle button does not add the badge in the parent album view. I believe you should be calling `albums.refresh()` to ensure a consistent view (like we do in `album.setPublic`).
* Why is there this redundant _Sensitive_ button at the very top of the album visibility dialog? I don't think it belongs in there (and certainly not so prominently; it should be at the bottom and probably should be a simple checkbox). Given that you added a nice, prominent button in the header, do we need it in the dialog box at all?

scripts/main/init.js
```diff
@@ -95,6 +97,7 @@ $(document).ready(function () {
 				}
 			}
 		})
+		.bind(["h"], album.toggle_nsfw_filter)
```
kamil4, line 100: One more thing: this needs a `return false;` to prevent the keystroke from being interpreted by the browser itself.

ildyria:
> * Why is there this redundant _Sensitive_ button at the very top of the album visibility dialog? I don't think it belongs in there (and certainly not so prominently; it should be at the bottom and probably should be a simple checkbox). Given that you added a nice, prominent button in the header, do we need it in the dialog box at all?

1. Because people might miss the triangle.
2. Just the triangle is not enough to give a proper explanation of what is going on. Especially with regard to the visible/hidden behavior.
3. The eye indicates the difference between:
    - not public (white)
    - public visible (green)
    - public hidden (orange)

   I was unsure how to integrate the cases:
    - not public & sensitive
    - public & sensitive
    - public hidden & sensitive
   
   This lead to the addition of the warning sign.
   
4. As the setting impact the visibility of the album depending of the default values set, I thought it would be a good idea to group it in the visibility pannel.

PR #235, opened 2020-12-24
Update README to remove ref to obsolete submodule

bastantoine:
Looks like since the v4 the Lychee main repo doesn't include the docs a submodule, so no need to keep it in the install steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,14 +30,11 @@ If you want to contribute and edit CSS or JS files, you need to rebuild Lychee. 
 # Clone Lychee
 git clone https://github.com/LycheeOrg/Lychee.git
 
-# Initialize the submodules (doc and frontend)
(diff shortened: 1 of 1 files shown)


ildyria:
The v4 does use a submodule. You can find it in `public/Lychee-front`

ildyria (approved):

bastantoine:
> The v4 does use a submodule. You can find it in `public/Lychee-front`

Indeed you’re right. My bad there was a typo in the commit message. I meant that the v4 doesn’t include the docs repo as a submodule, but not that there wasn’t any submodule used 😊.

ildyria:
By the way, feel free to join us on gitter: https://gitter.im/LycheeOrg/Lobby :)

PR #240, opened 2021-01-19
Various fixes

kamil4:
Assorted minor fixes:

* Don't show the "+" menu entry for smart albums in public mode.
* Clean up the NSFW button while fixing the above.
* Deleting the (content of) the Unsorted album was broken after the rename from `0` to `unsorted`.
* Many text fields didn't check for empty content before sending it to the server, resulting in ugly server errors. Even basic things like the login dialog. Whitespace-only entries generally trigger the same issue, hence the `trim` test.
* After adding/deleting/changing tags of photos, refresh the albums view if there are any tag albums (since they could be affected).

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -1011,7 +1011,7 @@ album.delete = function (albumIDs) {
 				});
 			} else if (visible.album()) {
 				albums.refresh();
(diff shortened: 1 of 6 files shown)


ildyria (approved):
Looks sane.

PR #242, opened 2021-01-25
Allow to share an album with users right from the album page

bastantoine:
This PR adds a new icon in the top bar, next to the album visibility option.
On click, it opens a modal with the list of users, asking to choose which one to share it with.

See LycheeOrg/Lychee#885 for the related code in the front

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -972,6 +972,7 @@ album.shareUsers = function (albumID, e) {
 		api.post("User::List", {}, (users) => {
 			$("#sharing_people_form").empty()
 			if (users !== {}) {
+				$("#sharing_people_form").append(`<p>${lychee.locale["SHARING_ALBUM_USERS_LONG_MESSAGE"]}</p>`);
 				$.each(users, (_, user) => {
 					$("#sharing_people_form").append(`<div class='choice'>
 						<label>
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -944,13 +944,14 @@ album.shareUsers = function (albumID, e) {
 		</form>`;
 
 		api.post("Sharing::List", {}, (data) => {
-			$("#sharing_people_form").empty();
-			if (data != undefined) {
-				if (data.users != undefined) {
-					$("#sharing_people_form").append(`<p>${lychee.locale["SHARING_ALBUM_USERS_LONG_MESSAGE"]}</p>`);
+			const sharingForm = $("#sharing_people_form");
+			sharingForm.empty();
+			if (data !== undefined) {
+				if (data.users !== undefined) {
+					sharingForm.append(`<p>${lychee.locale["SHARING_ALBUM_USERS_LONG_MESSAGE"]}</p>`);
 					// Fill with the list of users
 					data.users.forEach((user) => {
-						$("#sharing_people_form").append(lychee.html`<div class='choice'>
+						sharingForm.append(lychee.html`<div class='choice'>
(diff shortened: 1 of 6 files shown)


ildyria:
I will check that PR this Weekend.

scripts/main/album.js
```diff
@@ -937,6 +937,131 @@ album.setPublic = function (albumID, e) {
...
+	var usersNotToShare = [];
+	$(".basicModal .choice input").each((_, input) => {
+		if ($(input).is(":checked")) {
+			usersToShare.push(parseInt(input.name, 10));
+		} else {
+			usersNotToShare.push(parseInt(input.name, 10));
+		}
+	});
+
+	api.post("Sharing::List", {}, (data) => {
```
kamil4, line 1026: Didn't we invoke it earlier when opening the modal? Couldn't we cache the response from then? All you need to know now are the initial states and sharing `id`'s, correct? Couldn't you store them, say, in a `data-` attribute of each checkbox input?

kamil4 (changes requested):
I think it should be possible to simplify this; I left some comments to that effect. Although I will admit that you are restricted by the frankly ugly sharing API (that `Sharing::Delete`, for instance, is hair-raising; why can't it take the same arguments as `Sharing::Add`?! :worried:).

You also need to implement the hiding of the new header button in public mode, etc, in `header.setMode`. I'm guessing you should mimic the behavior of the album visibility button...

scripts/main/album.js
```diff
@@ -937,6 +937,131 @@ album.setPublic = function (albumID, e) {
...
+			if (data !== true) {
+				loadingBar.show("error", data.description);
+				lychee.error(null, params, data);
+			}
+		});
+	};
+
+	if (!basicModal.visible()) {
+		let msg = `<form id="sharing_people_form">
+			<p>${lychee.locale["WAIT_FETCH_DATA"]}</p>
```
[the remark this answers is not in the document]
bastantoine: When testing, I did see the message. It was only for a few seconds, but still.

I think it's better to have a message saying that it's normal if the user don't see anything yet, rather than displaying anything and just waiting to have the data from the API.

scripts/main/album.js
bastantoine: > All you need to know now are the initial states and sharing id's, correct?

Exactly. I did the API call twice, because I didn't figure out a way to keep the result of the call to `Sharing::List` when processing the choices of the user.

> Couldn't you store them, say, in a data- attribute of each checkbox input?

Yes I think it would work.

scripts/main/album.js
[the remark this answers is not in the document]
kamil4: Your server must be far away :smiley:.

scripts/main/view.js
```diff
@@ -493,13 +493,13 @@ view.album = {
 	},
 
 	public: function () {
-		$("#button_visibility_album").removeClass("active--not-hidden active--hidden");
```
kamil4, line 496: I don't think the changes to this file are needed at all. What this code does is it changes the _color_ of the "eye" button, depending on the public/hidden settings. See here:

https://github.com/bastantoine/Lychee-front/blob/d59325b793b834337a94f217b549a8ef6d628985/styles/main/_header.scss#L114-L120

But in case of your button, which is always gray, this won't do anything...

kamil4 (approved):
Looks good, apart from the seemingly unnecessary changes to one of the files?

PR #243, opened 2021-01-27
frontend code for disabling direction display on embedded map

sjml:
This is the frontend code addressing LycheeOrg/Lychee#886, reading the configuration object and not displaying the directional cone if the setting is off.

### DIFF
--- a/scripts/main/lychee.js
+++ b/scripts/main/lychee.js
@@ -200,6 +200,7 @@ lychee.init = function () {
 			lychee.image_overlay_type_default = lychee.image_overlay_type;
 			lychee.map_display = (data.config.map_display && data.config.map_display === "1") || false;
 			lychee.map_display_public = (data.config.map_display_public && data.config.map_display_public === "1") || false;
+			lychee.map_display_direction = (data.config.map_display_direction && data.config.map_display_direction === "1") || false;
 			lychee.map_provider = !data.config.map_provider ? "Wikimedia" : data.config.map_provider;
 			lychee.map_include_subalbums = (data.config.map_include_subalbums && data.config.map_include_subalbums === "1") || false;
 			lychee.location_decoding = (data.config.location_decoding && data.config.location_decoding === "1") || false;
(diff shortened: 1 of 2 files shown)


ildyria (changes requested):
This is fine but you are missing changes in 
https://github.com/LycheeOrg/Lychee-front/blob/master/scripts/main/lychee.js

Currently `lychee.map_display_direction` is not set.

sjml:
> This is fine but you are missing changes in
> https://github.com/LycheeOrg/Lychee-front/blob/master/scripts/main/lychee.js
> 
> Currently `lychee.map_display_direction` is not set.

Ah, thanks. I had thought that object was auto-populated from the database.

ildyria:
You may want to add something in line 200 and line 270 something in the `script/lychee.js`

sjml:
> You may want to add something near line 200 and line 270 in the `script/lychee.js`

Will do. I'll set it to default to `true`.

sjml:
This should all be set correctly now. @ildyria, look good to you?

ildyria (approved):
LGTM.

sjml:
Looking some more -- what about hiding the "Direction" data in the sidebar? I can do a push that accomplishes that, too, but could see it being a conceptually different thing (raw EXIF data vs. map visualization).

ildyria:
> Looking some more -- what about hiding the "Direction" data in the sidebar? I can do a push that accomplishes that, too, but could see it being a conceptually different thing (raw EXIF data vs. map visualization).

I don't have a strong opinion on that to be honest. My camera needs to be connected to my phone if I want to have GPS coordinates. Map visualization is very nice, but the use in practice makes it useless for me. I do see its use with phones but as a photographer, why would I take pictures with my phone when I can have better results with my full-frame camera. 

Then again in the case of phones, we need to trust the integrated compass which from experience is often off...

Conclusion, I don't care, maybe the other will have a stronger opinion.

By the way, feel free to drop on our gitter: https://gitter.im/LycheeOrg/Lobby you can easily login with your github ID. :)

PR #244, opened 2021-01-27
Fix rotation

kamil4:
This fixes image rotation (https://github.com/LycheeOrg/Lychee/issues/741).

The main fixes are on the server side. The front end changes do the following:
* Work with the new server rotation API that returns a full updated photo object rather than just `true`.
* Correctly handle the "2x" cases.
* Handle updates to album thumbs.
* Hide the rotation buttons for unsupported photo types.
* Update metadata in the sidebar.

There's also an unrelated small fix to the photo delete confirmation dialog removing a spurious space.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -1224,6 +1224,34 @@ album.isUploadable = function () {
 	return album.json.owner === lychee.username;
 };
 
+album.updatePhoto = function (data) {
+	if (album.json) {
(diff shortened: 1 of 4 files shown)


kamil4:
I think the remaining code smell suggestion is just silly so I have no plan to "fix" it.

ildyria:
I will review it this weekend.

ildyria (approved):

PR #246, opened 2021-02-13
Support upload_processing_limit

kamil4:
Implements a limit on the number of photos being simultaneously processed during upload.  I expected that adding this would complicate the `process` function but in fact it simplified it (since it was needlessly complicated to begin with).

This implements what was discussed in  https://github.com/LycheeOrg/Lychee/issues/903 and https://github.com/LycheeOrg/Lychee/issues/913. The default limit is `4`. Setting it to `0` disables the limit.

There will be a trivial corresponding server-side PR that adds a config variable.

### DIFF
--- a/scripts/main/lychee.js
+++ b/scripts/main/lychee.js
@@ -50,7 +50,7 @@ let lychee = {
 
 	album_subtitle_type: "oldstyle",
 
-	upload_processing_limit: 4,
+	upload_processing_limit: 2,
 
 	// this is device specific config, in this case default is Desktop.
 	header_auto_hide: true,
@@ -252,7 +252,7 @@ lychee.init = function () {
 			lychee.admin = !lychee.api_V2;
 			lychee.nsfw_visible_saved = lychee.nsfw_visible;
 
-			lychee.upload_processing_limit = parseInt(data.config.upload_processing_limit);
+			lychee.upload_processing_limit = parseInt(data.config.upload_processing_limit) || 4;
 
 			// leftMenu
 			leftMenu.build();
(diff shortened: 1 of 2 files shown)


ildyria (approved):
~~Not tested but looks okay~~.

~~Tested, BREAKING : you cannot upload more than one file at once.~~

Fixed.

pchampio:
Also fixes: https://github.com/LycheeOrg/Lychee/issues/443

scripts/main/lychee.js
```diff
@@ -50,7 +50,7 @@ let lychee = {
 
 	album_subtitle_type: "oldstyle",
 
-	upload_processing_limit: 4,
+	upload_processing_limit: 2,
```
kamil4, line 53: Why did you change it to `2`? And why only in this least important place?

scripts/main/lychee.js
```diff
@@ -252,7 +252,7 @@ lychee.init = function () {
 			lychee.admin = !lychee.api_V2;
 			lychee.nsfw_visible_saved = lychee.nsfw_visible;
 
-			lychee.upload_processing_limit = parseInt(data.config.upload_processing_limit);
+			lychee.upload_processing_limit = parseInt(data.config.upload_processing_limit) || 4;
```
kamil4, line 255: This is wrong. This prevents the value of `0` from being used. I don't think we can use this `||` syntax for numerals that can be `0`.

ildyria: Hummmm indeed, I did not think about this. The problem is that if you send `null` to `parseInt` you get `NaN` which obviously crash all the upload checks afterwards.

scripts/main/lychee.js
ildyria: I was testing. :'D

PR #249, opened 2021-02-16
Handle back button from albums view map correctly

kamil4:
Fixes https://github.com/LycheeOrg/Lychee/issues/917

### DIFF
--- a/scripts/main/header.js
+++ b/scripts/main/header.js
@@ -101,7 +101,7 @@ header.bind = function () {
 		lychee.goto(album.getID());
 	});
 	header.dom("#button_back_map").on(eventName, function () {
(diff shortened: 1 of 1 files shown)


d7415 (approved):
I haven't tested it, but I think that's sensible. I can't think of an edge case where an album would have ID 0?

kamil4:
> I can't think of an edge case where an album would have ID 0?

Well, _Unsorted_ used to, and still does sometimes on the server side I think, but in the front end that changed to `unsorted` when @ildyria refactored smart albums at one point.

Thanks for that comment though; it made me double-check that it works as expected with _Unsorted_ :smiley:.

ildyria (approved):

PR #253, opened 2021-03-03
add responsive web design for small screens

evoludolab:
This PR is an attempt at improving support for small screens in general and phones in particular. This adds responsive web design features to adjust the size of albums and photo thumbnails based on size of screen. It also adjusts margins and title size as well as suppresses subtitles to improve visibility of small thumbnails.

As an aside, I have some further ideas for improving the user experience on touch devices (for example, toggling the overlay on images doesn't seem to work for touch). Let me know if changes in this direction are of interest.

PR #254, opened 2021-03-06
improve support for touch devices

evoludolab:
Lychee's mouse focused interface is not always intuitive on touch devices. This PR attempts to improve several issues:

1. `:hover` does not work on touch devices and actually interferes with the user handling. Restricting CSS `:hover` to devices that actually support it resolves these issues (e.g. in order to open an album two single taps were required).
2. improvements to image overlays:

   1. title of the images in overlay (if visible) disappeared when moving the mouse (or tapping the screen) but the rest of the overlay remained visible. Now the overlay title is displayed regardless of whether the title/toolbar is visible.
   2. the display of overlays was toggled by mouse clicks, while the type of overlay could be changed using the `o`-key. To simplify the UI logic and make it accessible for touch devices, the two features are rolled into one: one more overlay mode 'none' is added that doesn't show an overlay. Now clicks, the `o`-key and taps all cycle through the different overlays available. As a consequence the setting `image_overlay` is no longer required (e.g. no overlay corresponds to `image_overlay_type === 'none'`) and was removed. This requires small changes in the back end, see LycheeOrg/Lychee#940.

3. increase size of some buttons for easier access on touch devices. in particular, the back button was easily missed on small devices; increase size of context menu entries.

Together with #253 this significantly improves the Lychee user experience on phones etc. I just mention again that unfortunately the two PRs most likely result in (easy) merge conflicts. If it helps, I have them merged on my machine and could lump them together into a single PR.

### DIFF
--- a/deps/basiccontext/scripts/basicContext.js
+++ b/deps/basiccontext/scripts/basicContext.js
@@ -123,10 +123,12 @@
 		let x = normalizedEvent.x,
 			y = normalizedEvent.y;
 
+		let container = document.querySelector(".basicContextContainer");
+
 		// Get size of browser
 		let browserSize = {
-			width: window.innerWidth,
-			height: window.innerHeight,
+			width: container.offsetWidth,
+			height: container.offsetHeight,
 		};
 
 		// Get size of context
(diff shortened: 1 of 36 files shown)


ildyria:
@evoludolab sorry for the delay, do you consider this PR ready for review ?

evoludolab:
Yes the PR is ready. Looking forward to your review.

ildyria:
@evoludolab can you fix the conflicts ? :)

PR #259, opened 2021-05-14
JSON response reports different size variants of the same photo more consistently

nagmat84:
See [PR #1003](https://github.com/LycheeOrg/Lychee/pull/1003) of repository [Lychee](https://github.com/LycheeOrg/Lychee) for details. This PR contains the necessary adaptions of the front-end to meet the re-factored JSON interface of the back-end.


scripts/frame/main.js
```diff
@@ -81,21 +81,21 @@ frame.next = function () {
 
 frame.refreshPicture = function () {
 	api.post("Photo::getRandom", {}, function (data) {
-		if (!data.url && !data.medium) {
+		if (!data.url && data.sizeVariants.medium === null) {
```
ildyria, line 84: before checking `data.sizeVariants.medium === null` you need to check that 
`data.sizeVariants !== null` Otherwise this code will throw an exception is `data.sizeVariants` is `null`. Yes, it is stupid I know. But better safe than sorry. :(

You can probably just do a check at the beginning if `sizeVariants` and `url` exists in order to avoid to propagate the checks down the line. ;)

scripts/main/album.js
```diff
@@ -1317,19 +1317,58 @@ album.updatePhoto = function (data) {
...
-				this.medium2x = data.medium2x;
-				this.medium2x_dim = data.medium2x_dim;
-				this.thumbUrl = data.thumbUrl;
-				this.thumb2x = data.thumb2x;
 				this.url = data.url;
 				this.size = data.size;
-
+				// Deep copy size variants
+				this.sizeVariants = {};
+				if(data.sizeVariants.thumb !== null) {
```
ildyria, line 1324: double check before executing this code that `data.sizeVariants` is not `null`.

ildyria (changes requested):
Looks quite sane. Just a small check to add to avoid sad crashes. :)

scripts/frame/main.js
nagmat84: Fixed with commit 119e753

scripts/main/album.js
nagmat84: Fixed with commit 119e753

ildyria (approved):

PR #262, opened 2021-05-24
Refactor timestamps

nagmat84:
This PR is related to [PR #1016](https://github.com/LycheeOrg/Lychee/pull/1016) for repository [Lychee](https://github.com/LycheeOrg/Lychee).

The frontend consumes iSO 8601 datetime strings (aka "JSON timestrings") from the REST API incl. correct timezone offsets and converts them to a localized string at the GUI layer for all datetime attribues.


ildyria (approved):
LGTM. (I actually read that PR before the back end one).

PR #263, opened 2021-05-25
Fix 1014

kamil4:
Fixes https://github.com/LycheeOrg/Lychee/issues/1014

Looks like it was broken by #259. The actual fix for that bug is in `build.js` but once I found that one and ran a grep or two my hair raised and I just had to keep going... So this also fixes missing smart album thumbs, missing thumbs in context menu lists (such as destination albums for move) and who knows what else :wink:.


ildyria (approved):
LGTM.

PR #265, opened 2021-06-01
Refactor timestamps anew

nagmat84:
This PR is a new attempt for the unfortunate, previous [PR #262](https://github.com/LycheeOrg/Lychee-front/pull/262).

Changes since the previous PR:

 - The method `lychee.locale.printDateTime` has been changed such that date/time strings are printed in localized format by with respect to the original timezone of the photo.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -660,7 +660,7 @@ album.setSorting = function (albumID) {
 				<option value='id'>` +
 		lychee.locale["SORT_PHOTO_SELECT_1"] +
 		`</option>
-				<option value='takestamp'>` +
+				<option value='taken_at'>` +
 		lychee.locale["SORT_PHOTO_SELECT_2"] +
 		`</option>
 				<option value='title'>` +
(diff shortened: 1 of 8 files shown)


kamil4 (approved):
This seems to work fine now; thanks!

One thing to point out is that the timezone is now always printed, resulting in a busier output. In particular, in the info sidebar, at least for me (I guess it's locale-dependent?) the timestamps now occupy two rows. I'm OK with that but I wonder if that changed will be universally loved...

nagmat84:
> One thing to point out is that the timezone is now always printed, resulting in a busier output. In particular, in the info sidebar, at least for me (I guess it's locale-dependent?) the timestamps now occupy two rows. I'm OK with that but I wonder if that changed will be universally loved...

It is not only locale-dependent but also depends on the resolution of the screen or more precisely the window size of your browser. If I use Firefox with a maximized window at 1920x1200px, then the timestamps tightly fit into a single line in the sidebar. If I decrease the window size just a small amount (about 10px), then the timestamps wrap into two lines. However, this also happens with other values if I reduce the window size even further.

PR #266, opened 2021-06-17
Rework scroll position saving

kamil4:
Fixes https://github.com/LycheeOrg/Lychee/issues/1027

This is essentially a cleanup of #230, which was restricted to albums with subalbums, as well as of some old code dating presumably to v3.

The bug in https://github.com/LycheeOrg/Lychee/issues/1027 was I believe due to a race condition between two callbacks with timeouts in `album.js`: one loading the content, the other restoring the scroll position. I solved that by moving the restoring code to `view.js` (where it gets invoked from the content loading callback).

In the process I noticed that there were two different versions of scroll position restoring code: one for the top level albums view (probably an old code dating back to v3), and another for albums with subalbums (from #230). We also didn't save the scroll position of albums without subalbums, which would trigger https://github.com/LycheeOrg/Lychee/issues/1027 as well. So I generalized the code to handle (hopefully) all cases: now the positions of all albums should be preserved, including the top level view and albums without subalbums.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -198,17 +198,6 @@ album.load = function (albumID, refresh = false) {
 			});
 		} else {
 			processData(data);
-			// save scroll position for this URL
-			if (data && data.albums && data.albums.length > 0) {
-				setTimeout(function () {
-					let urls = JSON.parse(localStorage.getItem("scroll"));
-					let urlWindow = window.location.href;
-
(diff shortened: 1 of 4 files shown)


kamil4:
OK, I made an additional tweak for search results but this should be ready for review now.

ildyria (approved):
LGTM

PR #267, opened 2021-07-05
New Photos Email Notification

micahhenshaw:
This is for the PR on the Backend #1049. Notes are on that PR

### DIFF
--- a/scripts/main/lychee.js
+++ b/scripts/main/lychee.js
@@ -75,6 +75,7 @@ let lychee = {
 	checkForUpdates: "1",
 	update_json: 0,
 	update_available: false,
+	new_photos_notification: true,
 	sortingPhotos: "",
 	sortingAlbums: "",
 	location: "",
@@ -144,6 +145,7 @@ lychee.init = function (exitview = true) {
 		lychee.update_json = data.update_json;
 		lychee.update_available = data.update_available;
 		lychee.landing_page_enable = (data.config.landing_page_enable && data.config.landing_page_enable === "1") || false;
+		lychee.new_photos_notification = false;
 
 		lychee.versionCode = data.config.version;
 		if (lychee.versionCode !== "") {
@@ -231,6 +233,8 @@ lychee.init = function (exitview = true) {
 			lychee.editor_enabled = (data.config.editor_enabled && data.config.editor_enabled === "1") || false;
 
 			lychee.nsfw_visible_saved = lychee.nsfw_visible;
+
+			lychee.new_photos_notification = (data.config.new_photos_notification && data.config.new_photos_notification === "1") || false;
 
 			lychee.upload_processing_limit = parseInt(data.config.upload_processing_limit);
 			// when null or any non stringified numeric value is sent from the server we get NaN.
--- a/scripts/main/lychee.js
+++ b/scripts/main/lychee.js
@@ -75,7 +75,7 @@ let lychee = {
 	checkForUpdates: "1",
 	update_json: 0,
 	update_available: false,
-	new_photos_notification: true,
+	new_photos_notification: false,
 	sortingPhotos: "",
 	sortingAlbums: "",
 	location: "",
--- a/scripts/main/lychee_locale.js
+++ b/scripts/main/lychee_locale.js
@@ -298,7 +298,8 @@ lychee.locale = {
 
 	NEW_PHOTOS_NOTIFICATION: "Send new photos notification emails.",
 	SETTINGS_SUCCESS_NEW_PHOTOS_NOTIFICATION: "New photos notification updated",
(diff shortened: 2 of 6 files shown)


scripts/main/lychee.js
```diff
@@ -75,6 +75,7 @@ let lychee = {
 	checkForUpdates: "1",
 	update_json: 0,
 	update_available: false,
+	new_photos_notification: true,
```
kamil4, line 78: I'm curious why this defaults to `true` given that it's supposed to be an opt-in feature?

scripts/main/lychee_locale.js
```diff
@@ -9,6 +9,7 @@ lychee.locale = {
 	SETTINGS: "Settings",
```
kamil4, line 9: I don't remember you updating the locale files on the server side? You'll have to add these strings to every language file (sorry!).

scripts/main/notifications.js
```diff
@@ -0,0 +1,30 @@
+let notifications = {
+	json: "",
+};
+
+notifications.update = function (params) {
+	if (params.email.length > 1) {
+		var regexp = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
kamil4, line 7: :astonished:

Ugh, it's past 11pm so my ability to parse crazy regexps is severely diminished. What is this _expected_ to pass? :smiley:

scripts/main/settings.js
```diff
@@ -521,6 +521,24 @@ settings.changeLocationShowPublic = function () {
...
+	if ($("#NewPhotosNotification:checked").length === 1) {
+		params.new_photos_notification = "1";
+	} else {
+		params.new_photos_notification = "0";
+	}
+	api.post("Settings::setNewPhotosNotification", params, function (data) {
+		if (data === true) {
+			loadingBar.show("success", lychee.locale["SETTINGS_SUCCESS_NEW_PHOTOS_NOTIFICATION"]);
+			lychee.new_photos_notification = params.new_photos_notification === "1";
+			lychee.new_photos_notification_saved = lychee.new_photos_notification;
```
kamil4, line 535: What is this line for?

scripts/main/view.js
```diff
@@ -1489,6 +1508,51 @@ view.full_settings = {
...
+		lychee.content.html('<div class="settings_view"></div>');
+	},
+
+	content: {
+		init: function () {
+			view.notifications.clearContent();
+
+			$(".settings_view").append(
+				'<div class="setting_line"><p>' +
+					"Add your email below to enable receiving email notifications. " +
```
kamil4, line 1533: So this will not be localized, correct?

kamil4 (approved):
I haven't tested it but it looks generally sane...

scripts/main/lychee.js
micahhenshaw: Yep, that should be false

scripts/main/lychee_locale.js
micahhenshaw: Alright, so just put entries in each language file, just in English?

scripts/main/notifications.js
micahhenshaw: haha, that will just validate for email format, so plain strings will not pass

scripts/main/view.js
micahhenshaw: Not currently, but it can be if you'd like?

scripts/main/lychee_locale.js
ildyria: In all language files. We have the CI that checks that every lines in the English language is at least found in the other languages file (as place-holders).

**We do not expect you to do the translations**. Just copy over the same line as in the `English.php` file. That way, other users can directly look at their language file and edit the wrong/missing translations rather than having to find the missing lines between the two files.

scripts/main/view.js
ildyria: That would be nice. :)

scripts/main/notifications.js
micahhenshaw: @d7415 Would you care to elaborate? What would you do?

d7415: It's definitely a better pattern than most I've seen, but there will be valid email addresses that it misses. And the fact that it's so close almost makes it worse. Simplest example being that it special-cases ipv4 addresses, but makes no allowance for ipv6 (which are presumably valid). But also I think that you can use multiple quoted sections in the local part (which no-one is likely to do, but the pattern allows for one, which people are almost as unlikely to use...).

It's not a major issue, just minimal benefit. It doesn't prevent undeliverable emails (e.g user@example.com or abcde@fghi.jkl or abc@999.999.999.999), so why bother?

PR #278, opened 2021-11-17
Fix menus

kamil4:
OK, here is the patch that fixes the menus that I would like us to merge before the next stable release. As usual, once I started poking around all sort of other minor irritations showed up so I included a few other fixes as well. But in general:

* Removed Import and New Album from smart albums (kept Upload since that one at least works)
* Fixed the code introduced in #253 that moves some of the button functions to the context menu on narrow screens. It added them unconditionally, even if the particular functionality shouldn't have been available (either due to the public mode or smart or tag albums)
* Removed cover image selection entry from the context menu for smart and tag albums since it doesn't work there
* NSFW button was not available for tag albums even though it does work there
* Import via GUI in albums view would often try to load non-existent album "0" when finished. I could've changed it to "unsorted" but I think it's better to simply reload the albums view since that's where the user was to begin with.
* Under unclear circumstances (race condition?), the info sidebar could be empty for albums. Turns out there was already code in place to handle that but a condition got reversed in #253, presumably accidentally.

### DIFF
--- a/scripts/main/contextMenu.js
+++ b/scripts/main/contextMenu.js
@@ -17,6 +17,9 @@ contextMenu.add = function (e) {
 
 	if (visible.albums()) {
 		items.push({ title: build.iconic("tags") + lychee.locale["NEW_TAG_ALBUM"], fn: album.addByTags });
+	} else if (album.isSmartID(album.getID())) {
+		// remove Import and New album if smart album
+		items.splice(1);
 	}
 
 	if (!lychee.admin) {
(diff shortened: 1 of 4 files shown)


ildyria (commented):
LGTM

scripts/main/contextMenu.js
```diff
@@ -27,38 +30,53 @@ contextMenu.add = function (e) {
...
-			fn: () => album.setNSFW(album.getID()),
-		});
+	if (visible.album() && album.isUploadable()) {
+		// prepend further buttons if menu bar is reduced on small screens
+		let albumID = album.getID();
+		if (album.isTagAlbum()) {
+			// For tag albums the context menu is normally not used.
+			items = [];
+		}
+		if (albumID !== "starred" && albumID !== "public" && albumID !== "recent") {
```
nagmat84, line 40: I wonder how stable this condition is with respect to future changes or additions. May gut feelings tells me that the condition would be more reliable, if it dit not exclude certain albums which don't support the actions but if it included albums which do support the actions explicitly.

Proposal: `if( Number.isInteger(albumID) || albumID === "unsorted" )`.

nagmat84 (approved):
One minor proposal for improvement, but generelly it looks good. I also have no strong opinion, if my proposal would really improve the quality.

scripts/main/contextMenu.js
kamil4: I like your suggestion though I needed to add a `parseInt` in there as well since `albumID` is always a string on `master` (I seem to remember that that might change with your refactoring though?).

PR #280, opened 2022-01-15
Markdown support

kamil4:
Fixes https://github.com/LycheeOrg/Lychee/issues/1189

This adds support for markdown in photo and album descriptions.
- I used the `marked` parser as it was the smallest actively supported one I could find (<50 KB).
- I modified the editing of descriptions to provide a multi-line `textarea` rather than a single-line `input`; this required some tweaks as textareas are poorly supported in basicModal and such.
- I added support for displaying album descriptions right under the header. This is enabled by default (though configurable) because I felt that not having it displayed anywhere was an omission in our UI, but I can be persuaded otherwise.
- Much of the work went on the CSS side since we simply didn't support many of the HTML tags that markdown uses in the contexts in which descriptions are displayed. I made sure that basic markdown syntax renders nicely, such as headings/bold/italic/blockquotes/lists/code/rules/links; no guarantees about anything more than that, though it should be easy to extend should the need arise.

There is a corresponding trivial server PR that adds two config variables, one for displaying the album descriptions, the other for enabling markdown.

The PR includes two unrelated fixes:
- The separate view mode was crapping out on opening the sidebar because `photo.updateSizeLivePhotoDuringAnimation` is not defined in it.
- If the last visited album had an info sidebar open, it would reopen when going to any of the Settings/Logs/Diagnostics/etc. This was because `view.photo.hide()` was being called unconditionally but it should only be called in photo view.


kamil4:
SonarCloud fails because CSS uses `font-family: revert` to request going back to browser-defaults for backquotes and such; I consider that to be a bug in SonarCloud, not in my code, though I suppose we could request some specific fonts instead...

styles/main/_content.scss
```diff
@@ -286,6 +286,80 @@
...
+			font-size: 20px;
+			line-height: 24px;
+			font-weight: bold;
+		}
+		h3 {
+			font-size: 16px;
+			font-weight: bold;
+		}
+		ol {
+			list-style: decimal inside;
```
nagmat84, line 335:
```suggestion
			list-style: decimal outside;
			padding-left: 40px;
```

`ol` and `ul` need some additional settings regarding a hanging indention on the left. Currently, subsequent lines of a multi line bullet point are not properly indented.

Actually, we need a left-padding (e.g. 40px is the default for Firefox) and the position of the bullet point should be within the padding, i.e. outside the principal box of the HTML element. Otherwise visual problems as the following will happen.

Screenshot_20220116_105542a

styles/main/_content.scss
```diff
@@ -286,6 +286,80 @@
...
+		}
+		ul {
+			list-style: disc inside;
```
nagmat84, line 338:
```suggestion
			list-style: disc outside;
			padding-left: 40px;
```

See previous comment.

styles/main/_imageview.scss
```diff
@@ -83,6 +83,79 @@
...
+				line-height: 29px;
+				font-weight: bold;
+			}
+			h3 {
+				font-size: 20px;
+				line-height: 24px;
+				font-weight: bold;
+			}
+			ol {
+				list-style: decimal inside;
```
nagmat84, line 131:
```suggestion
				list-style: decimal outside;
				padding-left: 40px;
```

styles/main/_imageview.scss
```diff
@@ -83,6 +83,79 @@
...
+			}
+			ul {
+				list-style: disc inside;
```
nagmat84, line 134:
```suggestion
				list-style: disc outside;
				padding-left: 40px;
```

styles/main/_sidebar.scss
```diff
@@ -208,6 +208,82 @@
...
+			line-height: 24px;
+			font-weight: bold;
+		}
+		h3 {
+			font-size: 16px;
+			line-height: 19px;
+			font-weight: bold;
+		}
+		ol {
+			list-style: decimal inside;
```
nagmat84, line 259:
```suggestion
			list-style: decimal outside;
			padding-left: 40px;
```

styles/main/_sidebar.scss
```diff
@@ -208,6 +208,82 @@
...
+		}
+		ul {
+			list-style: disc inside;
```
nagmat84, line 262:
```suggestion
			list-style: disc outside;
			padding-left: 40px;
```

nagmat84 (commented):
Overall, I like the PR. See my detailed comments for specific remarks.

Beside the problem with multi-line bullet points for which I have proposed a direct suggestion to the CSS, I noticed another problem if one uses code blocks. I did not take the time to find out which CSS is responsible, so I complain here ;-) in this general remark.

If a block of code is used, the gap between the labels and the content in the sidebar disappears. Moreover, the sidebar gets an ugly horizontal scroll bar.

This is how the good (normal) case looks like (without a code block):

Screenshot_20220116_105542a

This is how the bad case looks like (with a code block):

Screenshot_20220116_105955a

The block element with the code destroys the horizontal positioning. IMHO, the horizontal scroll bar should also not appear at the bottom of the side bar, but be within the code block if the code has too long lines. The CSS property should be `overflow-x: scroll`.

scripts/main/album.js
```diff
@@ -458,8 +458,9 @@ album.setTitle = function (albumIDs) {
 album.setDescription = function (albumID) {
 	let oldDescription = album.json.description ? album.json.description : "";
 
-	const action = function (data) {
-		let description = data.description ? data.description : null;
+	const action = function () {
+		let description = $("#album_desc").val();
+		if (description === "") description = null;
```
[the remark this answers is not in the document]
kamil4: Yes... I basically copied how the only other use of textarea in the frontend (custom CSS in settings) was doing it. But you are right that fixing basicModal would've been better. It doesn't look too difficult (famous last words), however, I notice that the front end is currently using a pinned version of basicModal that ignores any subsequent changes:

https://github.com/LycheeOrg/Lychee-front/blob/14fbba355b8f35bcca5fe7274081f5cf0df1c76e/package.json#L22

@ildyria, this is due to your commit 391a1d99a25fc0bd18a7dcb3472342d7a524e813 -- do you remember why you did it?

scripts/main/album.js
```diff
@@ -493,6 +494,7 @@ album.setDescription = function (albumID) {
 			},
 		},
 	});
+	$("#album_desc").html(oldDescription).select();
```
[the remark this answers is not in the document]
kamil4: I think it's because the value of a `textarea` doesn't really get interpreted as HTML; it is simply copied into the input box in a "row" form. I'll try your suggestion of using `.text` but I'm guessing that may not work correctly, actually (won't the escaped sequences become visible in the textarea box?).

nagmat84:
I have merged `consistent_json_api` into `master` and rebased your branch `markdown-support` as `markdown-support_rebased` onto the new master.
To update your PR, I believe the following commands should be the the right ones
**Attention**: I haven't tested them.

Commands:

```
git checkout master                                                      # Checkout the current master of Lychee-front, the only important thing is that your are not on the branch `markdown-support`
git branch -D markdown-support                                           # Delete your local branch `markdown-support`
git checkout -b markdown-support --track origin/markdown-support_rebased # Checkout a new local branch `markdown-support` from `markdown-support_rebased`
git pull                                                                 # Pull it
git branch -u origin/markdown-support                                    # Change upstream back to `markdown-support`
git push                                                                 # Try to push it, this is expected to fail, but make sure that GIT tries to push to the correct branch
git push --force                                                         # Now push forcefully
```

ildyria:
Better integration directly in [#1303](https://github.com/LycheeOrg/Lychee/pull/1303)

PR #284, opened 2022-03-03
Use filesize in original size variant - see Lychee PR #1239

Chostakovitch:
This PR goes with [this backend PR](https://github.com/LycheeOrg/Lychee/pull/1239) to adapt the migration of filesize from `Photo` to `SizeVariant`.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -1286,7 +1286,6 @@ album.updatePhoto = function (data) {
 	if (album.json) {
 		$.each(album.json.photos, function () {
 			if (this.id === data.id) {
-				this.filesize = data.filesize;
 				// Deep copy size variants
 				this.size_variants = {
 					thumb: null,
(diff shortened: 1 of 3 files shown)


scripts/main/album.js
```diff
@@ -1286,7 +1286,6 @@ album.updatePhoto = function (data) {
 	if (album.json) {
 		$.each(album.json.photos, function () {
 			if (this.id === data.id) {
-				this.filesize = data.filesize;
```
kamil4, line 1289: Shouldn't you update `filesize` in `deepCopySizeVariant` above?

kamil4 (changes requested):
Apart from the missing update in `album.updatePhoto`, it looks good!

scripts/main/album.js
Chostakovitch: Oh, nice catch, sorry for this.

nagmat84:
As the requested change by @kamil4 has been incorporated and the backend has already been, I am merging this one, too.

PR #286, opened 2022-04-09
Fix location type

nagmat84:
The attributes `latitude` and `longitude` are nullable JS numbers. The latitude and longitude 0°0'0''.0000 are valid coordinates, but falsy in JS. So we must not use shortcut notations like `if (latitude)` to check if we have a latitude/longitude, but explicitly check for `!== null` .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+npm
 node_modules/
 bower_components/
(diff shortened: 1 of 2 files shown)


ildyria (approved):
LGTM.

PR #288, opened 2022-04-11
Center clear search `X`

qwerty287:
Frontend part of https://github.com/LycheeOrg/Lychee/pull/1264

### DIFF
--- a/styles/main/_header.scss
+++ b/styles/main/_header.scss
@@ -192,7 +192,9 @@
 
 	&__clear {
 		position: absolute;
-		top: -2px;
+		top: 50%;
+		-ms-transform: translateY(-50%);
+		transform: translateY(-50%);
 		right: 8px;
 		padding: 0;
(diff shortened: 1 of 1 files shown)


ildyria (approved):

kamil4 (approved):
This looks good!

One thing that's still puzzling me is why in public mode (if you enable `public_search` in the config) the clear search button is positioned more to the left. Curiously, this is something that _I_ did as part of #110 -- clearly I was trying to fix some issue, but perhaps I was instead simply working around a browser bug that has since been fixed? Do you think we should remove this code now: https://github.com/LycheeOrg/Lychee-front/blob/feba79086476c7c522052b9b8d4c81b397a888c3/styles/main/_header.scss#L209-#L211

qwerty287:
I tested it, and if I remove those lines you mentioned, it looks like the non-public search. Should I just remove it to fix the issue?

PR #282, opened 2022-01-30
Consistent JSON API incl. Proper Handling of Errors (2nd half of big refactoring)

nagmat84:
# Summary

## API for access to Backend Revised

The API to access the backend has been revised and adopted to the new way how error and exceptions are reported. The frontend now properly observes the HTTP status code and only passes successful responses (2xx) to the success handler. If Lychee reports an exception, the received exception object is properly parsed and displayed to the user.

## Code for importing/uploaded got heavily revised

As the streamed response for a import of photos from the server has been changed drastically, the frontend code has been changed, too.
This is were most of the "real" work went into.

## Completely annotated JS code with parameter and returns types

This part has already been read by you and is responsible for most of the diff :-( although is does not change anything. Most of the TODOs which have been created are future work, because they are totally unrelated to the current re-factoring. Changes to the code were only made to the extent to make SonarCloud happy again. After adding the type annotation, the rating dropped to C because SonarCloud was suddenly able perform a much deeper static code analyse. But we are back to an A-rating now.
 
 
## Miscellaneous

Simplified and more secure handling of CSFR token.

PR #289, opened 2022-04-11
Fix untranslatable strings

qwerty287:
Frontend part of https://github.com/LycheeOrg/Lychee/pull/1270

### DIFF
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -53,7 +53,7 @@ api.post = function (fn, params, successCallback, responseProgressCB = null, err
 			if (isHandled) return;
 		}
 		// Call global error handler for unhandled errors
-		api.onError(api.isTimeout(errorThrown, jqXHR) ? "Session timed out." : "Server error or API not found.", params, errorThrown);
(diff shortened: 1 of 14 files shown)


ildyria:
While the intent is welcome, I fear that this one is not going to be merge, the reason being that #282 may revert some of those change.

I think a better approach would actually to make the PR on the [consistent_json_api](https://github.com/LycheeOrg/Lychee-front/tree/consistent_json_api) branch which will be merged. This is a way more future proof approach :)

ildyria:
Note that for https://github.com/LycheeOrg/Lychee/pull/1271/files it does not matter because those are not impacted by the big json refactoring.

qwerty287:
> I think a better approach would actually to make the PR on the consistent_json_api branch which will be merged. This is a way more future proof approach :)

So, I should close this and "recreate" it once #282 is merged?

ildyria:
yes, you can either wait or use [consistent_json_api](https://github.com/LycheeOrg/Lychee-front/tree/consistent_json_api) as a base and make PR to that branch instead of master. :)

qwerty287:
I'll close this and do it once again when #282 is finished.

PR #290, opened 2022-04-13
Add album tracks

qwerty287:
Adds a new feature that allows to upload GPX tracks/files and link them to an album (only regular, no smart/tag albums). This track is displayed together with the photos on the map, means this is only usable if the map is enabled.
To display the track, the Leaflet plugin https://github.com/mpetazzoni/leaflet-gpx is used, it is very easy and only needs one file.

Frontend part of https://github.com/LycheeOrg/Lychee/pull/1274, please also see the notes and explanations there.


scripts/main/contextMenu.js
```diff
@@ -79,6 +79,13 @@ contextMenu.add = function (e) {
 					});
 				}
 			}
+			if (!album.isSmartID(albumID) && lychee.map_display) {
+				// display track add button if it's a regular album
+				items.push({}, { title: build.iconic("location") + lychee.locale["UPLOAD_TRACK"], fn: () => $("#upload_track_file").click() });
```
nagmat84, line 84: Why does the event handler triggers another click-event? Why don't you directly call the relevant method `upload.uploadTrack`?

scripts/main/init.js
```diff
@@ -374,6 +378,10 @@ $(document).ready(function () {
 				}
 			}
 		);
+
+	// hide track uploader
+	$("#upload_track").hide();
```
nagmat84, line 383: Why is this necessary during the document-ready event?

scripts/main/mapview.js
```diff
@@ -281,6 +285,34 @@ mapview.open = function (albumID = null) {
 		updateZoom();
 	};
 
+	const setTrack = function (album) {
+		// add track
+		if (album.track_id) {
+			mapview.trackLayer = new L.GPX("uploads/tracks/" + album.track_id + ".xml", {
```
nagmat84, line 291: If there was `album.track_short_path`, then it would not be necessary to hard-code any file path here. It would simply be `L.GPX(album.track_short_path)`.

scripts/main/mapview.js
```diff
@@ -281,6 +285,34 @@ mapview.open = function (albumID = null) {
...
+					shadowUrl: null,
+				},
+			})
+				.on("error", function (e) {
+					lychee.error(lycche.locale["ERROR_GPX"] + e.err);
+				})
+				.on("loaded", function (e) {
+					if (
+						album.photos.filter((element) => {
+							return element.latitude || element.longitude;
```
nagmat84, line 305:
```suggestion
							return element.latitude !== null || element.longitude !== null;
```
Please keep in mind, that the geographically position 0°0'0.000'' / 0°0'0.000'' is very unlikely but valid.

scripts/main/mapview.js
```diff
@@ -281,6 +285,34 @@ mapview.open = function (albumID = null) {
 		updateZoom();
 	};
 
+	const setTrack = function (album) {
```
nagmat84, line 288:
```suggestion
	/**
	 * Add a track to the map.
	 *
	 * @param {(Album|PositionData)} album
	 */
	const setTrack = function (album) {
```

scripts/main/mapview.js
```diff
@@ -296,6 +328,7 @@ mapview.open = function (albumID = null) {
 		 */
 		const successHandler = function (data) {
 			addPhotosToMap(data);
+			setTrack(data);
```
nagmat84, line 331: I wonder if it is the right approach to have two methods here which both take the retrieved album as an argument. In particular, both methods do not only add content to the map, but zoom and pan the map such that the map covers the bounding box of the content. At least the last step is redundant, because it is executed by both methods.

IMHO, either we should combine everything into one method `addContentToMap` or we should have three methods and pull out the zooming step in an extra method, i.e. `addPhotosToMap`, `addTrackToMap`, `zoomToBoundingBox`.

scripts/main/upload.js
```diff
@@ -918,3 +918,220 @@ upload.check = function () {
 		$resync.prop("checked", false).prop("disabled", true);
 	}
 };
+
+/**
+ * @param {(FileList|File[])} files
+ */
+upload.uploadTrack = function (files) {
```
nagmat84, line 925: I believe this method can heavily be simplified as you only can add a single track to an album.

nagmat84 (changes requested):
Great work and a great feature. But still something to do ...

scripts/main/contextMenu.js
qwerty287: Before calling the method, a file chooser mist be shown (this is the same behavior as the image upload is doing it).

scripts/main/upload.js
qwerty287: I think I did this as far as possible, but most other methods (especially `upload.show`) are built for lists and require them. Most other usages of this list are as simple as possible.

scripts/main/mapview.js
qwerty287: I merged both methods into `addContentToMap`, an `updateZoom` was already there and I just updated it.

scripts/main/upload.js
kamil4: I'll take a crack at simplifying it further...

kamil4 (changes requested):

scripts/main/contextMenu.js
nagmat84: You are right. I don't like the frontend but that is nothing you could be blamed for. So we keep it that way for consistency reasons, although I hate it. I regard this conversation as resolved.

nagmat84 (changes requested):
I see a light at the end of the tunnel :grinning: First have a look at https://github.com/LycheeOrg/Lychee/pull/1274#discussion_r850690321

kamil4:
I am confused why the CI complains about the formatting when it finds no problems if I run it locally...

nagmat84:
It was not my intention to close this PR. As with the other PR for Lychee, this PR got closed when the base branch of this PR `consistent_json_api` was deleted. Same thing to do as for the Lychee part

- Edit PR and change base branch from `consistent_json_api` to `master`
- Checkout the rebased branch and push it here.

nagmat84 (changes requested):
Sorry, you are the first one to suffer from my newly found demand for proper and complete jsDoc comments. It really helps to improve static code analysis (of my IDE and also of SonarCloud).

I put a lot of (tedious) work in my recent re-factoring to improve the code in this respect and annotate all the existing code. I would like to try to hold that standard.

scripts/main/upload.js
nagmat84: @kamil4 Did you change anything?

nagmat84 (approved):
Tested and approved :+1:

scripts/main/upload.js
kamil4: "Changed" doesn't accurately capture the extent of what I did to this code. More like "decimated", really :wink:.

nagmat84: And still my IDE complains about 40 lines of duplicated code. But I did not want to take care of that. When the “Import from URL” will also use streamed responses one day, then we should clean that up. (dreaming)

PR #291, opened 2022-04-29
Add drag / drop

qwerty287:
Adds support to drag and drop albums and photos.

The only issue I found and can't fix: if you drop an album on the border of a smart album, the album is not recognized as smart album. I don't know any way to fix it.

Requested in https://github.com/LycheeOrg/Lychee/issues/922

### DIFF
--- a/scripts/main/contextMenu.js
+++ b/scripts/main/contextMenu.js
@@ -154,7 +154,7 @@ contextMenu.album = function (albumID, e) {
 
 /**
  * Handles drop event of an album onto an album and shows context menu to let the user pick the actions.
- * 
+ *
  * @param {string} sourceAlbumID source album (which is being dragged)
  * @param {string} targetAlbumID target album (where it is dropped)
  * @param {DragEvent} e


scripts/main/build.js
```diff
@@ -45,15 +45,17 @@ build.getAlbumThumb = function (data) {
 	var thumb2x = "";
 
 	if (thumb === "uploads/thumb/" && isVideo) {
-		return `<span class="thumbimg"><img src='img/play-icon.png' alt='Photo thumbnail' data-overlay='false' draggable='false'></span>`;
+		return `<span class="thumbimg"><img id='album-${data.id}' src='img/play-icon.png' alt='Photo thumbnail' data-overlay='false' draggable='false'></span>`;
```
kamil4, line 48: I don't think you need to add any of these `id`s to ``s. The information you need can be extracted from the ancestor (`album` or `photo` `<div>`).

scripts/main/build.js
```diff
@@ -101,9 +103,14 @@ build.album = function (data, disabled = false) {
 
 	let html = lychee.html`
 			<div class='album ${disabled ? `disabled` : ``} ${data.is_nsfw && lychee.nsfw_blur ? `blurred` : ``}'
+				id='album-${data.id}'
 				data-id='${data.id}'
```
kamil4, line 107: Same here. No need to add the `id` because the `<div>` already contains this info in `data-id` and in `class`.

scripts/main/build.js
```diff
@@ -101,9 +103,14 @@ build.album = function (data, disabled = false) {
...
 				data-nsfw='${data.is_nsfw ? `1` : `0`}'
-				data-tabindex='${tabindex.get_next_tab_index()}'>
+				data-tabindex='${tabindex.get_next_tab_index()}'
+				draggable='${album.isSmartID(data.id) ? "false" : "true"}'
+				ondragstart='startAlbumDrag(event)'
+				ondragover='event.preventDefault()'
+				ondrop='dropAlbum(event)'>
```
kamil4, line 113: Smart albums can be neither the source nor the destination for such dragging, so I think at least `ondrop` should be conditional as well? Also, it's not just smart albums but also the tag albums.

Also, it would be nice to provide a visual feedback when hovering over a destination album while dragging (I think the same visual effect that we have for regular hover would work fine).

scripts/main/build.js
```diff
@@ -227,9 +234,10 @@ build.photo = function (data, disabled = false) {
 	}
 
 	html += lychee.html`
-			<div class='photo ${disabled ? `disabled` : ``}' data-album-id='${data.album_id}' data-id='${
+			<div id='photo-${data.id}' class='photo ${disabled ? `disabled` : ``}' data-album-id='${data.album_id}' data-id='${
```
kamil4, line 237: Same here. Shouldn't be needed; this info is already provided.

scripts/main/dragDrop.js
```diff
@@ -0,0 +1,29 @@
+/**
```
kamil4, line 1: A general comment:

The front end tends to avoid global function names like this. Maybe have it as `dragdrop.start` and `dragdrop.finish` or something like that? Although, given how short this file is, I'm hesitating if it wouldn't make more sense to simply add these two functions to an existing file, such as `lychee.js`?

scripts/main/dragDrop.js
```diff
@@ -0,0 +1,29 @@
...
+	ev.dataTransfer.setData("text", ev.target.id);
+}
+
+/**
+ * drop album
+ * @param {DragEvent} ev
+ */
+function dropAlbum(ev) {
+	ev.preventDefault();
+	if (album.isSmartID(ev.target.id.substring(6))) return;
```
kamil4, line 15: Yeah, that's what I meant in an earlier comment about not providing `ondrop` for smart and tag albums in the first place -- I think that way you wouldn't need this test here at all?

As to your more general question/comment about it working weird for the corners or smart albums, it's not just the corners and not just the smart albums. Whatever the reason might be, `ev.target` is not necessarily `<div>` (or even ``), but if you drop, say, right over the album title, it will be `<h1>`. Adding `id` to all of them would be insane (which is part of the reason why I objected to it in the first place). But there should be an easy fix, something like:
```
    albumdiv = ev.target.closest("div.album");
```
and then you can again use `dataset.id` and `className`.

Also, you should check if the source and the destination are not the same album.

scripts/main/dragDrop.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * drag album to another one
+ * @param {DragEvent} ev
+ */
+function startAlbumDrag(ev) {
+	ev.dataTransfer.setData("text", ev.target.id);
```
kamil4, line 6: You should be able to extract what you need without defining `id`, using `dataset.id` and `className`, I guess?

kamil4 (changes requested):

ildyria:
I like that.

kamil4 (changes requested):
This looks much better! Other than the fact that, for me at least, it doesn't work at all out of the box (see one of the suggestions).

I still think the destination album should be highlighted while dragging over it. Is this something you are working on? I'll be happy to look into it myself if needed...

nagmat84:
I have merged `consistent_json_api` into `master` and rebased your branch `querty287/drag-drop` as `drag_drop_rebased` onto the new master. To update your PR, I believe the following commands should be the the right ones.

**Attention**:
 - I haven't tested them.
 - I assume that the default remote `origin` points to your own repository `git@github.com:qwerty287/Lychee-front.git`.

Commands:

```
git checkout master                                                  # Checkout the current master of Lychee-front, the only important thing is that your are not on the branch `drag-drop`
git remote add lychee-org git@github.com:LycheeOrg/Lychee-front.git  # Add the LycheeOrg-Repo as a remote unless you haven't yet
git branch -D drag-drop                                              # Delete your local branch `sass`
git checkout -b drag-drop --track lychee-org/drag-drop_rebased       # Checkout a new local branch `sass` from `sass_rebased`
git pull                                                             # Pull it
git branch -u origin/drag-drop                                       # Change to upstream to your repository, this assumes `origin` points to `qwerty287`
git push                                                             # Try to push it, this is expected to fail, but make sure that GIT tries to push to the correct remote
git push --force                                                     # Now push forcefully
```

kamil4:
I added visual feedback during drag over destination albums.

I also flipped the order of arguments to `contextMenu.albumDrop` (to make it consistent with the rest).

I'm satisfied with this PR at this point, but @nagmat84 should check it with his fancy IDE :smiley:.

kamil4 (approved):

nagmat84:
> I'm satisfied with this PR at this point, but @nagmat84 should check it with his fancy IDE :smiley:.

The way you put it, it sounds ridiculous. :cry:

scripts/main/build.js
```diff
@@ -251,7 +261,10 @@ build.photo = function (data, disabled = false) {
 	html += lychee.html`
 			<div class='photo ${disabled ? `disabled` : ``}' data-album-id='${data.album_id}' data-id='${
 		data.id
-	}' data-tabindex='${tabindex.get_next_tab_index()}'>
+	}' data-tabindex='${tabindex.get_next_tab_index()}'
+			draggable='true'
+			ondragstart='lychee.startDrag(event)'
```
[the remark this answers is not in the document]
nagmat84: Unfortunately, there are a lot of these kinds of errors, because our JS code is still not in a form which can be interpreted by the IDE in a reliable way. I am no expert on modern build chains for JS applications.

In the old times a JS application was mostly a bunch of files which were merged in the final step. This is the way how I am used to JS development and by co-incident this is also the way how the Gulpfile is written. Basically all JS files are merged together. Seemingly the build system dates to the same time when I turned my back to JS development.

As far as I understand this is not the way you would do it today. Instead one would define proper JS modules (keyword `module`) and then use these modules in other modules (via keyword `use`). In the build system one would not list the files individually, but the build process would collect all required files automatically.

This also seems the reason why the IDE is unhappy. For example, I get a lot of errors that `jQuery` aka `$` was not defined. (Of course it is, because we include the file `jquery.min.js` into the merge, but the IDE does not detect that.) I also get errors that an object of type `jQuery` cannot be assigned to an object of type `jQuery`. :smiley: The proper way would probably be to `use` jQuery.

I considered changing that and refactor out build system, but as I am a total noob in this respect and I could not find a quick solution, I dropped it again.

Long story short: For the moment we need to be selective about which IDE error we take seriously and which we can ignore. There are a lot of false positive. However, if anybody with more knowledge about JS and Gulp feels like fixing our build system and how we include modules into each other, I would be happy.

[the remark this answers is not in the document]
kamil4: > So we can ignore this for now and merge this PR?

I think another set of eyes should look over this PR before it's merged. I believe @nagmat84 said that he may be able to do it over the weekend?

scripts/main/contextMenu.js
```diff
@@ -152,6 +152,34 @@ contextMenu.album = function (albumID, e) {
 	basicContext.show(items, e.originalEvent, contextMenu.close);
 };
 
+/**
+ *
+ * @returns {void}
+ * @param sourceAlbumID source album (which is being dragged)
+ * @param targetAlbumID target album (where it is dropped)
+ * @param {DragEvent} e
+ */
```
nagmat84, line 161:
```suggestion
/**
 * Handles drop event of an album onto an album and shows context menu to let the user pick the actions.
 * 
 * @param {string} sourceAlbumID source album (which is being dragged)
 * @param {string} targetAlbumID target album (where it is dropped)
 * @param {DragEvent} e
 *
 * @returns {void}
 */
```

Mainly the type annotation `{string}` was missing.

scripts/main/lychee.js
```diff
@@ -1173,3 +1173,80 @@ lychee.getBaseUrl = function () {
...
+	/** @type string */
+	let type = "album";
+	/** @type ?Element */
+	let div = ev.target.closest("div.album");
+	if (!div) {
+		type = "photo";
+		div = ev.target.closest("div.photo");
+		if (!div) return;
+	}
+	ev.dataTransfer.setData("text", `${type}-${div.dataset.id}`);
```
nagmat84, line 1192:
```suggestion
	/** @type ?HTMLDivElement */
	const div = ev.target.closest("div.album,div.photo");
	if (!div) return;
	const type = div.classList.contains("album") ? "album" : "photo";
	ev.dataTransfer.setData("text/plain", `${type}-${div.dataset.id}`);
```

 - More efficient, because it only traverses the DOM tree once and does not potentially traverse up to the root in case the first search for `div.album` fails
 - `const` is always better than `let` ;-)
 - According to the official HTML specs, the type must be a proper MIME type. Actually we could even use `application/json` and transfer a JSON object.

scripts/main/lychee.js
```diff
@@ -1173,3 +1173,80 @@ lychee.getBaseUrl = function () {
...
+/**
+ * drop album
+ * @param {DragEvent} ev
+ * @returns {void}
+ */
+lychee.finishDrag = function (ev) {
+	ev.preventDefault();
+
+	/** @type string */
+	const data = ev.dataTransfer.getData("text");
```
nagmat84, line 1204:
```suggestion
	const data = ev.dataTransfer.getData("text/plain");
```

scripts/main/lychee.js
```diff
@@ -1173,3 +1173,80 @@ lychee.getBaseUrl = function () {
...
+};
+
+/**
+ * Album drag-over callback
+ * @param {DragEvent} ev
+ * @returns {void}
+ */
+lychee.overDrag = function (ev) {
+	ev.preventDefault();
+	/** @type ?Element */
```
nagmat84, line 1225:
```suggestion
	/** @type ?HTMLDivElement */
```

scripts/main/lychee.js
```diff
@@ -1173,3 +1173,80 @@ lychee.getBaseUrl = function () {
...
+};
+
+/**
+ * Album drag-leave callback
+ * @param {DragEvent} ev
+ * @returns {void}
+ */
+lychee.leaveDrag = function (ev) {
+	/** @type ?Element */
+	let div = ev.target.closest("div.album");
```
nagmat84, line 1239:
```suggestion
	/** @type ?HTMLDivElement */
	const div = ev.target.closest("div.album");
```

nagmat84: For `HTMLDivElement` see https://developer.mozilla.org/en-US/docs/Web/API/HTMLDivElement. I don't know if you know or use MDN, but this is my primary resource. Much better than any questionable tutorial or Stack Overflow.

And if you are more the hardcore guy, the ultimate truth is always: https://html.spec.whatwg.org/multipage/grouping-content.html#the-div-element. Luckily, MDN always links to the primary source, but I find MDN much more reader-friendly.

nagmat84 (changes requested):
Overall, looks good. I have added some improvements though. I hope I did not make a copy&paste error when I moved the changes from my code to suggestions here.

I tested it and it works. However, I discovered a bug which has probably introduced by my big refactoring. If you drag/drop albums onto other albums within a sub-album, the confirmation message always shows "Untitled" instead of the correct title of the album. I have a fix ready and will push it independently.

scripts/main/build.js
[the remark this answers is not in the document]
nagmat84: With respect to this point, yes. However, I made some other suggestions.

nagmat84 (approved):
IMHO, ready for merge

Issue #208, opened 2020-07-23
suggestion: remove google fonts dependency


guzzisti:
### Detailed description of the problem
There are imports of google fonts currently in page.css and landing.css.
I'd like to suggest to remove this dependency either by shipping the required fonts with the package or just avoid the usage of another font. 
Self-hosting a personal image gallery and make it depend on other services seems pointless for me.

As far as i can tell the page looks fine even without the additional font.

tmp-hallenser:
I think we don't use the fonts. Can someone confirm this? In the main.css delivered to the browser, I don't see any reference to roboto font.

guzzisti:
I've grep'ed through the files shipping with 4.0.6:
```
grep -r * -e "Roboto"
public/installer/assets/css/style.css:@import url("https://fonts.googleapis.com/css?family=Roboto:400,300,500,700,900");
public/installer/assets/css/style.css:	font-family: Roboto, Helvetica Neue, Helvetica, Arial, sans-serif;
public/installer/assets/css/style.css:	font-family: Roboto, Helvetica Neue, Helvetica, Arial, sans-serif;
public/installer/assets/css/style.css:	font-family: Roboto, Helvetica Neue, Helvetica, Arial, sans-serif;
public/installer/assets/css/style.css:	font-family: Roboto, Helvetica Neue, Helvetica, Arial, sans-serif;
public/installer/assets/css/style.css:	font-family: Roboto, Helvetica Neue, Helvetica, Arial, sans-serif;
public/installer/assets/css/style.css:	font-family: Roboto, Helvetica Neue, Helvetica, Arial, sans-serif;
public/installer/assets/css/style.css:	font-family: Roboto, Helvetica Neue, Helvetica, Arial, sans-serif;
public/dist/page.css:@import url("https://fonts.googleapis.com/css?family=Roboto:300,400,700,900");
public/dist/page.css:  font-family: 'Roboto', sans-serif;
public/dist/page.css:      font-family: 'Roboto', sans-serif;
public/dist/page.css:  font-family: 'Roboto', sans-serif;
public/dist/landing.css:@import url("https://fonts.googleapis.com/css?family=Roboto:300,400,700,900");
public/dist/landing.css:  font-family: 'Roboto', sans-serif;
public/dist/landing.css:      font-family: 'Roboto', sans-serif;
[...]

PR #294, opened 2022-05-05
Remove Google Fonts

qwerty287:
Serve fonts directly from local files.
Closes https://github.com/LycheeOrg/Lychee-front/issues/208

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ NO_COLOR="\033[0m"
 GREEN="\033[38;5;010m"
 YELLOW="\033[38;5;011m"
 
-echo "\n${GREEN}pre commit hook start${NO_COLOR}\n"
+printf "\n${GREEN}pre commit hook start${NO_COLOR}\n"
 
 PRETTIER="./node_modules/prettier/bin-prettier.js"
 
@@ -85,13 +85,13 @@ if [ -x "$PRETTIER" ]; then
     done
 else
     echo ""
-    echo "${YELLOW}Please install prettier, e.g.:${NO_COLOR}"
+    printf "${YELLOW}Please install prettier, e.g.:${NO_COLOR}"
     echo ""
     echo "  npm install"
     echo ""
 fi
 
-echo "\n${GREEN}pre commit hook finish${NO_COLOR}\n"
+printf "\n${GREEN}pre commit hook finish${NO_COLOR}\n"
 ```
 
 This can easily be installed by doing:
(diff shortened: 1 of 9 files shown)


ildyria:
I disagree with this last commit, if you enable the Landing page, roboto is being used.

ildyria:
> Do you mean we shoud keep Roboto there? Because I changed the font used on the landing page, it is not longer Roboto (take a look at CSS changes in this PR).

Yes I saw and tested. :)

With Roboto:
2022-05-06_1920x1080_11:50:57
2022-05-06_1920x1080_11:52:52

Without:
2022-05-06_1920x1080_11:52:56
2022-05-06_1920x1080_11:53:31

It is a personal opinion, but I quite preferred the thin look that Roboto gives compared to depending on the random Sans-serif font used by the browser (to be clear I am not opposed to hearing other opinions).

kamil4:
@ildyria I can certainly understand your preference in this case. However, the question is: _Should this be the default?_ In particular, if this effect can be achieved using our custom CSS feature (I haven't tried so I don't know), wouldn't that be the way to go for advanced users, rather than pushing non-standard fonts on everybody?

If the use of the custom CSS is problematic in this case, I'm certainly OK with using a (self-hosted) Roboto font on the landing page though, as an initial version of this PR did...

nagmat84:
What is the progress of this PR? I believe all arguments have been said. @qwerty287 has committed a user guidance on how to create a custom `user.css` which re-enables the Roboto font.

I feel rather indifferent about the decision. But I would like to see this PR merged soon, either as it is now (with no dependency on Roboto at all) or as it was before the last commit (with a self-hosted version of Roboto). How do we come to a decision?

nagmat84:
We could also take a ballot among the active developers. As we are five (@ildyria, @kamil4, @d7415, @qwerty287 and me) we will get a definitive decision, if no one abstains.

styles/landing/landing.scss
```diff
@@ -1,5 +1,28 @@
 // Rest -------------------------------------------------------------------- //
-@import url("https://fonts.googleapis.com/css?family=Roboto:300,400,700,900");
+/* roboto-300 - vietnamese_latin-ext_latin_greek-ext_greek_cyrillic-ext_cyrillic */
+@font-face {
+	font-family: "Roboto";
+	font-style: normal;
+	font-weight: 300;
+	src: local(""), url("../../../fonts/roboto-v29-300.woff2") format("woff2");
```
kamil4, line 7: Just curious -- why does this need `"../../../"`? What directory are we in when these need to be loaded? I can see that it's correct from the point of view of the Lychee front end sources, but once those are compiled, doesn't that whole nested folder structure get forgotten?

kamil4 (approved):
Approved, assuming it works (see my question about the relative paths).

PR #295, opened 2022-05-05
Mobile improvements

qwerty287:
Various improvements to the UI, mainly addresses mobile users.

### DIFF
--- a/styles/main/_sidebar.scss
+++ b/styles/main/_sidebar.scss
@@ -243,7 +243,7 @@
 @media only screen and (max-width: 567px), only screen and (max-width: 640px) and (orientation: portrait) {
 	// sidebar as overlay, small size
 	.sidebar {
-		width: 240px;
+		width: 300px;
 		height: unset;
 		background-color: rgba(0, 0, 0, 0.6);
 
@@ -286,7 +286,7 @@
 
 		#tags {
 			margin: 4px 0 6px 10px;
-			width: calc(100% - 16px);
+			width: 284px;
 
 			.empty {
 				margin: 0;


styles/main/_content.scss
```diff
@@ -297,8 +297,8 @@
 
 		.album,
 		.photo {
-			// 3 thumbnails per row
-			--size: calc((100vw - 3px) / 3);
+			// 2 thumbnails per row
+			--size: calc((100vw - 5px) / 2);
```
nagmat84, line 301: That's something we need to discuss and I have to test on various smartphones, resp. For example, I am very happy with 3 photos per row on my phone and 2 photos only would be a waste of space.

Maybe we need to be more specific about the the intended screen width, at the moment  3 thumbnails per row is used for devices with a screen width between 320px and 567px. Maybe we have to use more cases here, i.e. use 2 thumbnails per row for 320px...440px and 3 thumbnails for 440px...567px. (Note, I just invented the 440px for the example, it is not an informed choice.)

styles/main/_logs_diagnostics.scss
```diff
@@ -12,7 +12,8 @@
 		-moz-user-select: text;
 		-ms-user-select: text;
 		user-select: text;
-		width: fit-content;
+		overflow-x: scroll;
+		overflow-y: hidden;
```
nagmat84, line 16: Why don't you want the user to scroll down to older log messages?

Moreover, `<pre>` is used in a block-formatting context and is a block-level element, i.e. its intrinsic height always matches the content (unless we have changed that somewhere, I need to check that) This means the CSS property `overflow-y` won't have any effect unless `height` is specified, too. (Again I need to check first, if we do so somewhere.)

nagmat84 (commented):
The CSS looks fine in general, nothing to say against it on the level of code quality. I have to check first with various devices, whether I like the switch to a 2-two-photo-per-row layout.

PR #296, opened 2022-05-05
Fix regression bug

nagmat84:
During review of #291 I noticed that `album.buildMessage` does not work as expected, because the previous version only searches for the designated album ID in `albums.json.albums`, i.e. inside the root album. Hence, the confirmation message in #291 did not show the correct title of albums when used inside sub-albums.

I cannot remember to have changed this method during my re-factoring, but it seems so. Now it should be fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
 		"gulp-load-plugins": "^1.5.0",
 		"gulp-sass": "^5.1.0",
 		"gulp-uglify": "^3.0.2",
-		"npm": "^6.14.16",
+		"npm": "^6.14.17",
 		"prettier": "2.2.1",
 		"sass": "^1.51.0",
 		"uglify-js": "^3.12.4"
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -1147,36 +1147,30 @@ album.getArchive = function (albumIDs) {
  * @returns {string} the HTML content of the dialog
  */
 album.buildMessage = function (albumIDs, albumID, op1, op2, ops) {
-	let title = "";
-	let sTitle = "";
+	let targetTitle = lychee.locale["UNTITLED"];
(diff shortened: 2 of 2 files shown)


ildyria (approved):

kamil4 (approved):
Yeah, I noticed this the other day but I couldn't remember if it was actually that way before the big merge.

I'm pretty sure that we don't handle _all_ the possible cases (and never have). Things used to be more reliable in that respect back when `albums.json` contained a complete tree structure, but of course `albums.json` may be reset to `null` after many edit operations and then we are back to square 1...

Passing just the album IDs and trying to recover the titles from them is never going to work reliably given how our front end works. We should really be passing the titles themselves, because the caller will pretty much always have access to that info. Given the uncertain future prospects of the current front end though, I can't be bothered to fix that.

nagmat84:
> Yeah, I noticed this the other day but I couldn't remember if it was actually that way before the big merge.

Me neither. But the bug was so obvious to me that I simply assumed that it has to be a regression bug.
 
> Passing just the album IDs [...] is never going to work reliably [...]. We should really be passing the titles themselves, because the caller will pretty much always have access to that info.

Or even better: Pass the album/photo/whatever object itself. If one really wanted to go over the edge, we would even attach the JS object to the DOM when the frontend builds the HTML structure for an album/photo listing. At the moment we simply use the `data`-attribute of a `HTMLElement` to set some string ID. The DOM is not limited to adding custom string attributes. (Only our build function is.) In an ideal world we would link our "business" object (i.e. our models) to the HTML element and have it ready in our event handlers and wherever we need them. We would not even need to search for them and as JS does not make deep copies, but stores references this would be super-efficient.

> Given the uncertain future prospects of the current front end though, I can't be bothered to fix that.

Me, too.

PR #297, opened 2022-05-06
Fix coloring in scripts

qwerty287:
Must use `printf` instead of `echo`. Tested with bash and zsh.


d7415 (approved):
I haven't tested the hook, but my shell's echo/printf behave as described.

PR #300, opened 2022-06-12
Support API token per user

qwerty287:
Frontend for https://github.com/LycheeOrg/Lychee/pull/1368

### DIFF
--- a/scripts/3rd-party/backend.js
+++ b/scripts/3rd-party/backend.js
@@ -211,7 +211,6 @@ const SmartAlbumID = Object.freeze({
  * @property {?string} email
  * @property {boolean} may_upload
  * @property {boolean} is_locked
- * @property {string} token
  * @property {boolean} has_token
  */
 


kamil4 (approved):
LGTM (untested).

scripts/main/settings.js
```diff
@@ -483,3 +483,47 @@ settings.save_enter = function (e) {
 		},
 	});
 };
+
+settings.viewToken = function () {
+	api.post("User::getCurrent", {}, function (data) {
+		let bodyHtml = "";
+		let enableReset = "";
+		if (data.token === "") {
+			bodyHtml = "<div class='directLinks'><p>disabled</p></div>";
```
kamil4, line 492: Shouldn't this be localized?

scripts/main/settings.js
```diff
@@ -483,3 +483,47 @@ settings.save_enter = function (e) {
 		},
 	});
 };
+
+settings.viewToken = function () {
+	api.post("User::getCurrent", {}, function (data) {
```
nagmat84, line 488: Please annotate `data` with the correct jsDoc type as done everywhere else.

scripts/main/settings.js
```diff
@@ -483,3 +483,47 @@ settings.save_enter = function (e) {
...
+		let bodyHtml = "";
+		let enableReset = "";
+		if (data.token === "") {
+			bodyHtml = `<div class='directLinks'><p>${lychee.locale["DISABLED"]}</p></div>`;
+			enableReset = lychee.locale["ENABLE"];
```
nagmat84, line 493: What is `enableReset`?

scripts/main/settings.js
```diff
@@ -483,3 +483,47 @@ settings.save_enter = function (e) {
...
+		}
+		basicModal.show({
+			body: bodyHtml,
+			buttons: {
+				action: {
+					title: enableReset,
+					fn: function () {
+						api.post("User::resetToken", {});
+						basicModal.close();
+					},
```
nagmat84, line 511: I don't understand what is supposed to happen here. According to my understanding `resetToken` regenerates the token and returns a new token, correct?

But I don't see that you do something with the result which is returned by `resetToken`. If you essentially throw it away, why not `unsetToken`?

nagmat84 (changes requested):
I recommend to have a look at the review of the backend first. Here are some thoughts beyond those in the comments

 - Please have look into [/scripts/3rd-party/backend.js](https://github.com/LycheeOrg/Lychee-front/blob/06e068debb5ffb67734d5b563e79aeb82fef0dbc/scripts/3rd-party/backend.js). I did a great deal in adding jsDoc tags to all methods etc. to get at least some kind of type safety for editors which support it. Your methods are missing that.
 - I have to revoke my initial question from the review of the backend whether the token is a public information. (I first assumed that it would be some kind of public key for a private key-) However, I have learned that it is nothing more than a symmetric, shared secret between the backend and the frontend. Hence, it should also treated as one and handled like a password. In particular, it should not be possible to query the current token from the backend again. One should only be able to see it once after it has been generated. If the user forgets the token, then the user must regenerate a new token.
 - In principle, I don't have anything again adding a API call to get the authenticated user. However, here it seems unnecessary and only produces an unnecessary roundtrip. I always wanted to add the currently authenticated user as a proper object to the return value of the `init` call because we need information about the current user (like the username) all the time. See my TODO comments in the `SessionController::init()`. I don't ask you to to all the refactoring now, but maybe this is the time to at least add the user object.

scripts/main/settings.js
nagmat84: Then it should be named like that, i.e. `buttonCaption` or `submitButtonCaption`, or, or, ...

scripts/main/settings.js
nagmat84: This sounds wry. I would have expected to immediately see the result of a call to `resetToken`. I assume I have to check this PR our and test it myself how it works.

But, if we are never going to use the result of `resetToken`, because we need to query for the authenticated user upon the next time the dialog is visited anyway, why does `reset Token` return something at all?

scripts/main/settings.js
nagmat84: It is possible. I did that at a lot of places. You simply need to write the annotation in front of the lambda between the preceding comma `,` and the keyword `function` like this

```js
api.post("User::getCurrent", {}, /** here goes the annotation */ function (data) {
```

For better readability you can use line breaks

```js
api.post(
	"User::getCurrent",
	{},
	/**
	 * here goes the annotation
	 */
	function (data) {
```

Or the third alternative is to pull out the lambda, define it as a proper method like

```js
/**
 * here goes the annotation
 */
const getAuthenticatedUserHandler = function (data) {
}
```

But not annotating the parameter is not an option for me.

scripts/main/settings.js
nagmat84: Ah, I now see I have already done that. :+1:

scripts/main/view.js
```diff
@@ -1054,6 +1054,9 @@ view.settings = {
 		 * @returns {void}
 		 */
 		setLogin: function () {
+			const hasToken = true;
+			const tokenBtnText = lychee.locale[hasToken ? "RESET_TOKEN" : "ENABLE_TOKEN"];
```
nagmat84, line 1058: `hasToken` is always true here.

If you don't feel like fixing it, I would at least recommend to remove this and also remove `ENABLE_TOKEN` from the locales as is never used right now.

I better approach would be as follows which mostly requires changes in the backend.

In `App\Models\User`

 * add `token` to the `$hidden` array
 * add `has_token` to the `$appends` array
 * write an accessor method `getHasTokenAttribute(): bool` which basically returns `$this->token !== null`

As the object of the authenticated user is nowadays sent to the frontend, you can then use `lychee.user.has_token` here in the frontend.

scripts/main/settings.js
```diff
@@ -488,3 +488,50 @@ settings.save_enter = function (e) {
...
+						fn: basicModal.close,
+					},
+				},
+			});
+			$("#button_copy_token").on(lychee.getEventName(), function () {
+				navigator.clipboard.writeText(data.token);
+			});
+			$("#button_disable_token").on(lychee.getEventName(), function () {
+				api.post("User::unsetToken", {}, function () {
+					$("#apiToken").html("disabled");
```
nagmat84, line 532: Should this not be

```suggestion
					$("#apiToken").html(lychee.locale["DISABLED"]);
```

? Otherwise the locale key `DISABLED` is never used and I stumbled across this, because I tried the German translation and was wondering, why this one isn't translated.

nagmat84 (changes requested):
I have two minor change requests below which should be solved.

Except from that, I am not 100% happy with this PR, but I would deem it okay-ish and fine for merge. (Mostly, because I feel too tired right now to implement it myself.) Nonetheless let me sketch how I would like the PR even more. But this is really only the sugar on the cream and not a "must".

Currently, the UX feels somehow strange.

 * On the main page of the settings, there is the button "Reset API token" which opens the dialog, but also resets the token at the same time. Hence, there is no way to simply check whether a token has been set or not. If users just want to ensure that no token is set, they first have to open the dialog (and therewith inevitably create a token) just to press the delete button within the dialog afterwards.
 *  The dialog itself has another "Reset" button. If one presses this button, the dialog shortly vanishes and re-opens with a new token. This causes a strange "flicker" which disturbs me mostly. Moreover, users might be wondering, why the button is labeled differently than the button on the main page of the settings even though both buttons do exactly the same.

This is how I would like it more:

* On the main page of the settings, the button should only be labeled "API token ..." (the three dots indicate that another dialog will be opened)
* If one presses the button, only the dialog will be opened, nothing more. In particular it should not reset the token.
* The dialog only has only a single close button at the bottom, not a second "Reset" button in addition.
* The two buttons on the right are extended by a third button to reset the token, the icon could look like the one below
* The string which shows the token value should support three values:
    * `Disabled`: Just like now. Should also be shown directly after the dialog has been opened for users with `lychee.user.has_token === false`.
    * The actual token value: Should be shown after the user has pressed the newly added third button the reset the token
    * `Enabled`: Should be shown directly after the dialog has been opened for users with `lychee.user.has_token === true`. As we cannot show a previously set token, this would only indicate an existing token and increase symmetry with the value `Disabled`
 * If users press the button to reset the token, the value of the token inside the dialog should be changed instantly like it already happens for deleting a token right now without closing and re-opening the dialog again.

arrow-rotate-left-solid

nagmat84:
I thought this would make it clear

> Except from that, I am not 100% happy with this PR, but I would deem it okay-ish and fine for merge. (Mostly, because I feel too tired right now to implement it myself.) Nonetheless let me sketch how I would like the PR even more. But this is really only the sugar on the cream and not a "must".

So it is basically up to you. Of course, I would like to see it fixed, but as I am the only one which seems to be bothered by that and I am not using that feature, it is also fine to not do it.

nagmat84:
If you have already done it half way, I would appreciate it if you could go the rest of the way, too. Typically, you one expects the buttons at the bottom to close the dialog, not to re-open it.

scripts/main/settings.js
```diff
@@ -492,46 +492,79 @@ settings.save_enter = function (e) {
...
+			}
+		);
+	});
+
+	$("#button_disable_token").on(lychee.getEventName(), function () {
+		api.post("User::unsetToken", {}, function () {
+			$("#button_copy_token").hide();
+			$("#button_disable_token").hide();
+			$("#apiToken").html(lychee.locale["DISABLED"]);
+			lychee.user.has_token = false;
```
nagmat84, line 567: You have already defined the method `updateBtnVisibility` above. Why don't you use it?

I'll take care of that with the other comment.

scripts/main/settings.js
```diff
@@ -492,46 +492,79 @@ settings.save_enter = function (e) {
...
-				});
-			});
+settings.openTokenDialog = function () {
+	let token = "";
+
+	/**
+	 * @param {boolean} has
+	 * @param {boolean} canBeViewed
+	 */
+	const updateBtnVisibility = function (has, canBeViewed) {
```
nagmat84, line 502: I'm not quite happy with the parameters.

`has` is to unspecific for my taste. I know what you mean from context, it means `hasToken` but I doubt that anybody does remember that in half a year. Moreover, the parameter is actually not necessary. The method can simply use `lychee.user.has_token` internally. No need, to pass that as an argument at all.

`canBeViewed` should rather be named like `isTokenValueVisible`. Again, in half a year nobody remembers _what_ can be viewed if the parameter is only called like like that.

I will change it.

nagmat84 (changes requested):
Some more change request, but I will do them myself.

scripts/main/settings.js
```diff
@@ -492,46 +492,79 @@ settings.save_enter = function (e) {
...
+	});
+
+	$("#button_disable_token").on(lychee.getEventName(), function () {
+		api.post("User::unsetToken", {}, function () {
+			$("#button_copy_token").hide();
+			$("#button_disable_token").hide();
+			$("#apiToken").html(lychee.locale["DISABLED"]);
+			lychee.user.has_token = false;
+		});
+	});
```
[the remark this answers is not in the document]
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/300/commits/bca3d5558ddcabeb52334f6397f425b1a549636d. IMHO this can be resolved.

scripts/main/settings.js
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/300/commits/bca3d5558ddcabeb52334f6397f425b1a549636d. IMHO this can be resolved.

scripts/main/settings.js
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/300/commits/bca3d5558ddcabeb52334f6397f425b1a549636d. IMHO this can be resolved.

nagmat84 (approved):
I am happy now. :smile: Only the remaining issues of the backend needs to be resolved.

PR #302, opened 2022-06-16
Add "Clear" button to logs

qwerty287:


kamil4 (commented):
Sorry if I missed a discussion on this, so I might be simply missing context. Is this supposed to completely clear the logs, whereas "clean noise" removes only debug messages and such? I feel like this might be too fine a distinction for the end users -- I worry we will forever get questions about the difference between "clean" and "clear"? Do we really need both? I feel like we should possibly keep just one (and that may very well be the new "Clear").

ildyria (approved):
LGTM.

PR #303, opened 2022-06-24
Fix tag albums

kamil4:
Fixes https://github.com/LycheeOrg/Lychee/issues/1307

Also fixes uploads via keyboard (`u`) and the confirmation string displayed for tag albums (which incorrectly informed that the contents of a tag album would be deleted).

Also includes an unrelated fix to prevent the info sidebar from being displayed in settings and such (if it was being displayed for an album and somebody then moved to the albums view and opened the settings).

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -1221,8 +1221,9 @@ album.delete = function (albumIDs) {
 		msg = `<p>` + lychee.locale["DELETE_UNSORTED_CONFIRM"] + `</p>`;
 	} else if (albumIDs.length === 1) {
 		let albumTitle = "";
+		const isTagAlbum = albums.isTagAlbum(albumIDs[0]);
 
-		action.title = lychee.locale["DELETE_ALBUM_QUESTION"];
+		action.title = lychee.locale[isTagAlbum ? "DELETE_TAG_ALBUM_QUESTION" : "DELETE_ALBUM_QUESTION"];
 		cancel.title = lychee.locale["KEEP_ALBUM"];
 
 		// Get title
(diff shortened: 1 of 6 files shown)


scripts/main/album.js
```diff
@@ -1221,8 +1221,9 @@ album.delete = function (albumIDs) {
 		msg = `<p>` + lychee.locale["DELETE_UNSORTED_CONFIRM"] + `</p>`;
 	} else if (albumIDs.length === 1) {
 		let albumTitle = "";
+		const isTagAlbum = albums.isTagAlbum(albumIDs[0]);
 
-		action.title = lychee.locale["DELETE_ALBUM_QUESTION"];
+		action.title = lychee.locale[isTagAlbum ? "DELETE_TAG_ALBUM_QUESTION" : "DELETE_ALBUM_QUESTION"];
```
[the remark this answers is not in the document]
d7415: This only applies to dingle albums. The `DELETE_ALBUMS_QUESTION` is used below for multi-select.

scripts/main/contextMenu.js
```diff
@@ -110,21 +110,21 @@ contextMenu.album = function (albumID, e) {
 	// Show merge-item when there's more than one album
 	// Commented out because it doesn't consider subalbums or shared albums.
 	// let showMerge = (albums.json && albums.json.albums && Object.keys(albums.json.albums).length>1);
-	const showMerge = true;
+	const showMergeMove = !albums.isTagAlbum(albumID);
```
nagmat84, line 113: I assume the comment above this line needs some updating, too.

scripts/main/contextMenu.js
```diff
@@ -198,29 +198,29 @@ contextMenu.albumMulti = function (albumIDs, e) {
 	// Show merge-item when there's more than one album
 	// Commented out because it doesn't consider subalbums or shared albums.
 	// let showMerge = (albums.json && albums.json.albums && Object.keys(albums.json.albums).length>1);
-	const showMerge = true;
+	const showMergeMove = albumIDs.every((albumID) => !albums.isTagAlbum(albumID));
```
nagmat84, line 201: See previous comment.

nagmat84 (approved):
LGTM. I would like to see two comments to be updated, too, such that future developers won't be confused, but still approved.

scripts/main/contextMenu.js
kamil4: Honestly, I'm inclined to simply remove the commented out code plus its associated comments. That "feature" is never coming back...

scripts/main/album.js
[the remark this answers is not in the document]
kamil4: You are both correct :smiley:. Multi-select is handled separately, and I chose not to deal with it, because the message could only be correct if one selects only tag albums or only regular albums, but not a mix of both. I decided that improving the most common case was "good enough"...

scripts/main/contextMenu.js
nagmat84: Fine for me. Better no comment than a comment which raises more questions than it answers.

Issue #65, opened 2019-02-25
Searched Image // Back Button Behaviour wrong 


SerenaButler:
When clicking an Image within a search result, the back button at the image page should point (back) to the searched image results, **NOT** the Album of the image.

kamil4:
I looked into it and I'm cautiously optimistic that it may be easier to solve than it seems. The search results are internally a (temporary) album. Right now that album is discarded when switching to the photo view (i.e., clicking on an image). But if we keep it and reprogram the behavior of the Back button, it should work. I'm sure there will be all sorts of tricky corner cases when a user tries to edit the results, but the basic use case should be doable...

alex-phillips:
Any chance this could get looked at? Would love this feature when browsing hundreds of photos organized by tags

PR #306, opened 2022-07-25
Support the back button in search results

### LINKED ISSUE #920: The "BACK" arrow button does not go back to the place I was before
### Detailed description of the problem [REQUIRED]
Generally, I was searching for some pictures in my image library.
But when I tried to go back, it did not go back to the place I was before (i.e. search results).

I think I should be able to go back where I was before, not to the parent gallery of the one I entered.
If I wanted to go to the specific gallery, I would find it by name (search result told me how it's named) and enter it manually.

That's my opinion : ) I was just confused how it works, but maybe it's normal.

**Steps to reproduce the behavior:**
1. I've searched for "nawi" (it gives me 3 results - two 2nd level galleries and one 1st level gallery with these two 2nd level inside)
2. Now I entered the one of them
3. Next I wanted to go back using the arrow far on the left
4. And... I am not back to the search results, but in the 1st level album of the 2nd level gallery I entered before

**Screenshots**
1) firefox_2021-02-14_10-40-42
2) firefox_2021-02-14_10-42-06
3) firefox_2021-02-14_10-42-29

### Output of the diagnostics  [REQUIRED]
Not needed.

### Browser and system
**OS:** Windows 7 PRO
**Browser:** 84.0.2 (32 bity)

kamil4:
Fixes https://github.com/LycheeOrg/Lychee/issues/920, https://github.com/LycheeOrg/Lychee/issues/921, #65.

This basically improves the functioning of the back button in the front end for search results. Currently when one clicks on an image in search results and then clicks the back button, the front end goes to the album containing the image, not back to the search results. Same with albums in search results. This PR fixes it as follows:
* The ID of the search results album is changed to `search/search-term`
* All the hardcoded uses of the `search` string in the front end are fixed
* `album.load()` gains a functionality of altering the `parent_id` of a newly loaded album; this enables the going back to search results. This mimics the support that was already there for images (required for smart and tag albums). For albums there are limitations though; if the user opens an album from search results and then navigates to its subalbum, the altered parent album gets lost and clicking the back button twice will not go back to the search results. Also, I know that @nagmat84 wasn't fond of changing the `parent_id` and would instead prefer if we added (and used) an `effective_partent_id` or such; I wanted to keep this PR small at least for the first round, but this might be a good opportunity to fix that if we think it's worth it.
* The largest changes by line count are to `lychee.load()`, which needed to be refactored to accommodate the more album-like behavior of the search results.
* `visible.search()` is fixed so that it does not return `true` when viewing, say, an image from the search results.

One thing I didn't change was that the search results are considered a kind of albums view (`visible.albums()` returns `true` when viewing the search results). It's nasty but it's been that way for so long that I don't dare to touch it in this PR; there are probably countless places in the front end that depend on it...

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -20,7 +20,7 @@ album.isSmartID = function (id) {
  * @returns {boolean}
  */
 album.isSearchID = function (id) {
-	return id === SearchAlbumID;
+	return id === SearchAlbumIDPrefix || id.startsWith(SearchAlbumIDPrefix + "/");
 };
 
 /**
@@ -50,7 +50,7 @@ album.getID = function () {
 
 	// this is a Lambda
 	let isID = (_id) => {
-		return album.isSmartID(_id) || /*album.isSearchID(_id) || */ album.isModelID(_id);
+		return album.isSmartID(_id) || album.isSearchID(_id) || album.isModelID(_id);
 	};
 
 	if (photo.json) id = photo.json.album_id;
@@ -176,15 +176,23 @@ album.deleteSubByID = function (albumID) {
 /**
  * @param {string} albumID
  * @param {?AlbumLoadedCB} [albumLoadedCB=null]
+ * @param {?string} parentID
  *
  * @returns {void}
  */
-album.load = function (albumID, albumLoadedCB = null) {
+album.load = function (albumID, albumLoadedCB = null, parentID = null) {
 	/**
 	 * @param {Album} data
 	 */
 	const processAlbum = function (data) {
 		album.json = data;
+
+		if (parentID !== null) {
+			// Used with search so that the back button sends back to the
+			// search results.
+			album.json.original_parent_id = album.json.parent_id;
+			album.json.parent_id = parentID;
+		}
 
 		if (albumLoadedCB === null) {
 			lychee.animate(lychee.content, "contentZoomOut");
--- a/scripts/main/contextMenu.js
+++ b/scripts/main/contextMenu.js
@@ -380,7 +380,7 @@ contextMenu.photo = function (photoID, e) {
 		{ title: build.iconic("trash") + lychee.locale["DELETE"], fn: () => photo.delete([photoID]) },
 		{ title: build.iconic("cloud-download") + lychee.locale["DOWNLOAD"], fn: () => photo.getArchive([photoID]) },
 	];
-	if (album.isSmartID(album.getID()) || album.isSearchID(album.getID) || album.isTagAlbum()) {
+	if (album.isSmartID(album.getID()) || album.isSearchID(album.getID()) || album.isTagAlbum()) {
 		// Cover setting not supported for smart or tag albums and search results.
 		items.splice(2, 1);
 	}
--- a/scripts/main/header.js
+++ b/scripts/main/header.js
@@ -138,7 +138,7 @@ header.bind = function () {
 
 	header.dom(".header__search").on("keyup click", function () {
 		if ($(this).val().length > 0) {
-			lychee.goto("search/" + encodeURIComponent($(this).val()));
+			lychee.goto(SearchAlbumIDPrefix + "/" + encodeURIComponent($(this).val()));
 		} else if (search.json !== null) {
 			search.reset();
 		}
--- a/scripts/main/lychee.js
+++ b/scripts/main/lychee.js
@@ -589,7 +589,10 @@ lychee.reloadIfLegacyIDs = function (albumID, photoID, autoplay) {
 lychee.load = function (autoplay = true) {
 	let hash = document.location.hash.replace("#", "").split("/");
 	let albumID = hash[0];
-	let photoID = hash[1];
+	if (albumID === SearchAlbumIDPrefix && hash.length > 1) {
+		albumID += "/" + hash[1];
+	}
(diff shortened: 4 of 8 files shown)


nagmat84 (commented):
I reviewed the code and it looks good. I still need to test it.

After I had inspected `lychee.js` in my IDE, I also gained an understanding on why we now have an index 2 in `lychee.load`. This method is spaghetti-code, but I know that you are not responsible for it.

Nonetheless it might be helpful to have a comment at the beginning of the method which gives an overview about the different cases which may occur. I assume we now have the following cases:

 - `/`  --> load root album, assign `null` to `albumID` and `photoID`
 - `/{albumID}` --> load the album, `photoID` equals given ID, assign `null` to `photoID`
 - `/{albumID}/{photoID}` --> load album and then the corresponding photo, assign the respective values to `albumID` and `photoID`
 - `/search/{term}` --> load or go back to "search" album for the given term, assign `' search/{term}'` as fictitious `albumID` and assign `null` to `photoID`
 - `/search/{term}/{photoID}` --> load photo within fictitious search album, assign `' search/{term}' ` as fictitious `albumID` and assign the given ID to `photoID`

Right?

kamil4:
> `lychee.load`. This method is spaghetti-code, but I know that you are not responsible for it.

I like to think that I actually simplified it a little but still, I agree that it is what I believe you like to call a "god" method: it does everything.

FYI, a previous version of this PR used the `-` character to separate the `search` prefix and the search string, but that actually required _more_ changes than keeping `/` (unfortunate as it may be). So in the current rewrite I decided to keep it.
 
> Nonetheless it might be helpful to have a comment at the beginning of the method which gives an overview about the different cases which may occur.

Good point. I'll add it.

nagmat84:
I played around with this PR and there is at least one regression: Login as admin, open any album, do a right-click on any photo, the normal context menu of the browser appears instead of the Lychee-specific context menu. The context menu still works for multiple selected photos.

Moreover, I noticed the following annoying UX experience. But this already exists on master, so it is no regression:

- I would expect to see the back button in the search album, not the cog wheel. The back button should behave like clearing the search.
- The "x" in the search edit field is oddly misplaced for me. It is ways too low and nearly out of the edit box.

nagmat84:
And there is some other inconvenience, but I do not consider this a blocker for the PR.

Create an album, e.g. with title "Find me" and upload a photo into the album. Type in "Find me" into the search field. Click on the found album "Find me", let the album open, click on the photo inside the album, let the photo open, go back two times into the "search album" again. The edit box will only show a prefix of the search term, e.g. "Fi".

I assume it has something to do with the timeout between consecutive searches while the user is typing. Depending on the typing speed, the preserved prefix varies in length. The faster you type, the longer the prefix gets. But I never managed to get the whole string back. Obviously, I cannot type fast enough :D Of course, it would be nice to preserve the latest search term, but if it is too difficult to accomplish I am fine with the current behavior. The current behavior is already a big improvement.

nagmat84 (approved):
I tested it and I confirm the found regression is gone. I did not find any other bugs or regressions. Just the little inconvenience about incomplete prefixes of search terms when one goes back to the search album. But this is no stopper for me.

kamil4:
That misplaced `x` is mysterious. I fixed it at least once in the past, and I feel like @qwerty287 may have recently fixed it as well. It seems to be browser-dependent; it currently looks good for me both in Firefox and in Chrome so I was hoping that it was fixed for good. What browser are you using, @nagmat84?

nagmat84:
www-client/firefox-103.0.2 on Gentoo

kamil4:
> Type in "Find me" into the search field. Click on the found album "Find me", let the album open, click on the photo inside the album, let the photo open, go back two times into the "search album" again. The edit box will only show a prefix of the search term, e.g. "Fi".

Can you verify if that's fixed now?

I also pushed a fix for an unrelated issue that shows up when deleting photos in photo view; normally the "next" photo in the album is shown then. If that next photo was a video file, however, it would auto-play, but it shouldn't, since the user didn't click on it. That should be fixed now.

> www-client/firefox-103.0.2 on Gentoo

Same here, and it looks fine for me with my regular profile, but when I tweak the DPI settings, I see the problem as well -- I'll try to figure it out for good, but possibly not in this PR.

> I would expect to see the back button in the search album, not the cog wheel. The back button should behave like clearing the search.

This goes back to the following comment I made it the first post:

> One thing I didn't change was that the search results are considered a kind of albums view (visible.albums() returns true when viewing the search results). It's nasty but it's been that way for so long that I don't dare to touch it in this PR; there are probably countless places in the front end that depend on it...

So yeah, I agree that search results should be a (variant of) album view, which would result in the behavior you advocate, but they currently aren't, they've never been, and I would prefer cleaning that up to be a separate future PR...

nagmat84:
> > Type in "Find me" into the search field. Click on the found album "Find me", let the album open, click on the photo inside the album, let the photo open, go back two times into the "search album" again. The edit box will only show a prefix of the search term, e.g. "Fi".
> 
> Can you verify if that's fixed now?

I do confirm :+1:

> > I would expect to see the back button in the search album, not the cog wheel. The back button should behave like clearing the search.
> 
> This goes back to the following comment I made it the first post: [...]
> 
> So yeah, I agree that search results should be a (variant of) album view, which would result in the behavior you advocate, but they currently aren't, they've never been, and I would prefer cleaning that up to be a separate future PR...

Fine for me.

PR #301, opened 2022-06-13
Fix untranslatable strings

qwerty287:
Frontend of https://github.com/LycheeOrg/Lychee/pull/1371

PR #307, opened 2022-07-25
Fix 1419

### LINKED ISSUE #1419: Photo View link does not work
### Detailed description of the problem [REQUIRED]
When I copy the link to Lychee Photo View, it just display the title of the photo, not the photo itself.

**Steps to reproduce the behavior:**
1. Go to any photo in any public album
2. Click on Share button (the symbol), select Direct Links and copy the link to Lychee Photo View
3. Paste the link in a browser.
4. See error, i.e just the title of the photo is displayed. I also noticed that the info button and share button (even when album) not work. More: the Share button is visible even when "Share button is visible" is unchecked in the album share settings.

**Screenshots**
No screenshot, but see
https://photos.nikodemuskarlsson.se/view?p=QGY3AJYLenj4NMqO8tscV68X
(it should be a photo on the page too).

### Output of the diagnostics  [REQUIRED]
* https://photos.nikodemuskarlsson.se/Diagnostics

### Browser and system
 I run Safari v.15.5, Lychee v. 4.5.2 from Git Master 89fcba4 and PHP 8.1.8.

kamil4:
Fixes https://github.com/LycheeOrg/Lychee/issues/1419

* As usual, the standalone view mode got out of sync with the rest of the front end, especially after the #301 got merged.
* The view mode also didn't respect the `is_share_button_visible` setting.
* The info sidebar didn't work. I fixed it in #280 already but since that one got stalled, I'm moving the fix here.
* Unrelated to https://github.com/LycheeOrg/Lychee/issues/1419, I noticed that drag/drop wasn't disabled for read-only albums (in particular, in public mode) so that's now fixed as well.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -30,7 +30,12 @@ paths.view = {
 		"./scripts/3rd-party/backend.js",
 		"./deps/basiccontext/scripts/basicContext.js",
 	],
-	scripts: ["node_modules/jquery/dist/jquery.min.js", "node_modules/lazysizes/lazysizes.min.js", "../dist/_view--javascript.js"],
+	scripts: [
+		"node_modules/jquery/dist/jquery.min.js",
+		"node_modules/lazysizes/lazysizes.min.js",
+		"node_modules/sprintf-js/dist/sprintf.min.js",
+		"../dist/_view--javascript.js",
+	],
 	svg: ["./images/iconic.svg", "./images/ionicons.svg"],
 };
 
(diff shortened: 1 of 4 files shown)


qwerty287 (approved):
Does localization work here?

ildyria (approved):
LGTM.

kamil4:
> Does localization work here?

As far as I know, the standalone photo view is not localized. It doesn't seem to call `Session::init` at all, which is what enables our localization to work...

PR #308, opened 2022-07-27
support multiple paths, space separated

ildyria:
See https://github.com/LycheeOrg/Lychee/pull/1153

### DIFF
--- a/scripts/main/upload.js
+++ b/scripts/main/upload.js
@@ -511,7 +511,7 @@ upload.start = {
 
 		/**
 		 * @typedef ServerImportDialogResult
-		 * @property {string} path
+		 * @property {string} paths
 		 * @property {boolean} delete_imported
 		 * @property {boolean} import_via_symlink
 		 * @property {boolean} skip_duplicates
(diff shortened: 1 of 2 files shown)


scripts/main/upload.js
```diff
@@ -520,12 +520,13 @@ upload.start = {
...
 		const action = function (data) {
-			if (!data.path.trim()) {
-				basicModal.error("path");
+			if (!data.paths.trim()) {
+				basicModal.error("paths");
 				return;
 			} else {
 				// Consolidate `data` before we close the modal dialog
 				// TODO: We should fix the modal dialog to properly return the values of all input fields, incl. check boxes
+				data.paths = data.paths.match(/(?:\\.|\S)+/g);
```
nagmat84, line 529: Why this complicated Rexex? We want to split at every space except for spaces which are escaped by `\`, right? So why do we need the question mark and the colon? Shouldn't `/(\\ |\S+)+/g` be sufficient? I haven't tested it yet.

In the long run I would rather prefer to have a model dialog which dynamically creates another input field when the existing input fields are non-empty and which removes an input field again, if its input is deleted. This means there should always be a list of input field with one empty input field at the bottom. This way we could avoid using fragile regular expressions at all.

nagmat84 (commented):
LGTM in general, not tested. One question about the regex.

scripts/main/upload.js
ildyria: > why do we need the question mark and the colon?

`( ... )` defines a capturing section
`(?: ... )` defines a non capturing section.

nagmat84: I haven't the time to test it and obviously I am too tired to grasp the correctness of this expression. I assume you have tested it and it does what it is supposed to do? For me it still looks wrong even after I know that `(?: ...) defines a non-capturing group.

ildyria: https://stackoverflow.com/a/66319988/3981163

ildyria:
https://github.com/LycheeOrg/Lychee/pull/1153

nagmat84 (approved):
I approve this PR, because it is an improvement over the current situation.

However, while I have been working on a completely unrelated PR (more fine granular sharing), I noticed that importing more than a single file from Dropbox at once is fundamentally broken. Yes, it is only "somewhat" related to this PR, because this PR considers multiple paths to be imported from the server (not Dropbox). Nonetheless, the whole matter needs an overhaul.

PR #310, opened 2022-08-05
Add user and better structure to session json

nagmat84:
This are the front-end changes for https://github.com/LycheeOrg/Lychee/pull/1443.

### DIFF
--- a/scripts/3rd-party/backend.js
+++ b/scripts/3rd-party/backend.js
@@ -276,14 +276,11 @@ const SmartAlbumID = Object.freeze({
 /**
  * @typedef InitializationData
  *
- * @property {number} status - `1`: unauthenticated, `2`: authenticated
- * @property {boolean} admin
- * @property {boolean} may_upload
- * @property {boolean} is_locked
+ * @property {?User} user
+ * @property {{is_admin: boolean, is_locked: boolean, may_upload: boolean}} rights
  * @property {number} update_json - version number of latest available update
  * @property {boolean} update_available
  * @property {Object.<string, string>} locale
- * @property {string} [username] - only if user is not the admin; TODO: Change that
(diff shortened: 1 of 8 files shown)


ildyria (approved):
LGTM.

kamil4 (approved):

nagmat84:
BTW, this is not ready for merge if it is merged after https://github.com/LycheeOrg/Lychee/pull/1403 which I expect. In this case, this PR needs further adoptions, because the object `lychee.user` needs to be updated after user-updating requests. https://github.com/LycheeOrg/Lychee/pull/1403 introduces a change in how users are updated which this PR is not prepared for.

nagmat84:
After https://github.com/LycheeOrg/Lychee/pull/1403 has been merged into the backend, this PR needed to adopt. So I think it should be re-reviewed. The following aspects have been changed (or actually added):

* `settings.createLogin` updates `lychee.user`
* `settings.changeLogin` updates `lychee.user`

This has been necessary, because https://github.com/LycheeOrg/Lychee/pull/1403 now updates the session of the backend after a user has been changed, and I want to keep the frontend synced. Technically, this is not necessary _now_, because there is currently no code which relies on condition that the backend and frontend user are the same, but I want to avoid future surprises.

kamil4 (approved):

ildyria (commented):
LGTM.

Untested.

ildyria:
I was looking at the return on `Session::init()`:
```
rights: {is_admin: true, is_locked: false, may_upload: true}
user: {id: 0, username: "admin", email: "test@test.com", may_upload: true, is_locked: false}
```
I am wondering if we want this duplication of _rights_, or if we just don't care in this specific case. :)

I would say that the serialization in `toArray()` is needed because of the user administration page.

PR #311, opened 2022-08-11
Misc fixes

kamil4:
This was supposed to be three unrelated fixes but (almost by magic) one of the fixes ended up fixing two bugs, so it's down to two fixes:
- `photo.toggleStar` wasn't updating the parent album, so toggling the star status in photo view would not be immediately reflected when going back to the parent album. This is a regression from 4.4.0,
- The rotation buttons were being shown for non-editable albums (such as public or shared albums belonging to other users, while not being an admin). This required a larger fix because we were being sloppy with hiding these buttons if the photo was actually a video file and such. Now we unconditionally call `view.photo.show` when loading a new photo (even if already in photo view, as is the case with previous/next buttons), which ensures that the header buttons are in a well-defined state. This almost accidentally fixed another obscure bug I noticed, where for albums that are not downloadable and have "original" disabled, using previous/next between a video file and an image file would end up showing the "..." menu button for the image file, even though it should be hidden in that case.

### DIFF
--- a/scripts/main/header.js
+++ b/scripts/main/header.js
@@ -482,11 +482,11 @@ header.setMode = function (mode) {
 			}
 
 			if (album.isUploadable()) {
-				const e = $("#button_trash, #button_move, #button_visibility, #button_star");
+				const e = $("#button_trash, #button_move, #button_visibility, #button_star, #button_rotate_cwise, #button_rotate_ccwise");
 				e.show();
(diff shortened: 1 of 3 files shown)


ildyria:
Tested.

ildyria (approved):
LGTM.

ildyria:
IMHO, this can be merged immediately and then will be released with 4.6.0

PR #316, opened 2022-08-21
Bugfix for boolean condition in `isSearchID`

nagmat84:
Another victim of https://github.com/LycheeOrg/Lychee-front/pull/306. :grinning: I noticed that because I couldn't use "Import from Server ..." on the root level anymore and got `Uncaught TypeError: id is null` exception on the JS console.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -20,7 +20,7 @@ album.isSmartID = function (id) {
  * @returns {boolean}
  */
 album.isSearchID = function (id) {
-	return id === SearchAlbumIDPrefix || id.startsWith(SearchAlbumIDPrefix + "/");
+	return id !== null && id.startsWith(SearchAlbumIDPrefix);
 };
 
 /**
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -20,7 +20,7 @@ album.isSmartID = function (id) {
  * @returns {boolean}
  */
 album.isSearchID = function (id) {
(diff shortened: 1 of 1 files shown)


qwerty287 (approved):

scripts/main/album.js
```diff
@@ -20,7 +20,7 @@ album.isSmartID = function (id) {
  * @returns {boolean}
  */
 album.isSearchID = function (id) {
-	return id === SearchAlbumIDPrefix || id.startsWith(SearchAlbumIDPrefix + "/");
+	return id !== null && id.startsWith(SearchAlbumIDPrefix);
```
kamil4, line 23:
```suggestion
	return id !== null && (id === SearchAlbumIDPrefix || id.startsWith(SearchAlbumIDPrefix + "/"));
```

There's a reason why I coded it that way in the first place :smiley:. Imagine a regular album ID that happens to start with the string `search` -- not very probable, but far from impossible...

kamil4 (changes requested):

scripts/main/album.js
nagmat84: Okay ..... I actually assumed so, but than considered that we have 64 equally random symbols (a-z, A-Z, 0-9, -, _) and "search" has 6 characters, hence the chance is 1/64^6 ≈ 1/10^11, i.e. one time in more than a billion.

But, yes, better safe than sorry.

kamil4 (approved):

PR #318, opened 2022-08-28
Localization

nagmat84:
See https://github.com/LycheeOrg/Lychee/pull/1494.

PR #322, opened 2022-09-15
Adopt to new modern dialog

nagmat84:
This PR makes the Lychee frontend use the revised modal dialog (already merged https://github.com/LycheeOrg/basicModal/pull/3), with additional bug fixes by https://github.com/LycheeOrg/basicModal/pull/5.

### The advantages of using the revised modal dialog are:

* **Proper support for text areas** (Also a prerequisite for https://github.com/LycheeOrg/Lychee-front/pull/280).

* **Proper support for check boxes:** Their boolean values are now also part of the return object provided by the modal dialog, no need to use inefficient query selectors like in https://github.com/LycheeOrg/Lychee-front/blob/ab0f314cac720b28fbe1689a482bdd3af0f18efe/scripts/main/album.js#L826-L833

 * **Correct values types:** Strings from input fields, numbers from input fields, booleans from checkboxes, selections from - well - selections are returned with the proper JS type.

### The additional achievements of this PR on top are:

 * **Getting rid of `lychee.html`:** Modal dialogs don't need the unfortunate, home-brewed escape mechanism provided by `lychee.html` as in https://github.com/LycheeOrg/Lychee-front/blob/ab0f314cac720b28fbe1689a482bdd3af0f18efe/scripts/main/album.js#L382. Instead the property `.textContent` is used after the dialog has been created to set the text inside an element which allows to use full Unicode as JS strings are UTF-16. This ensures that no non-ASCII character in any user-provided string (title, description, etc.) can trigger errors. This is also a prerequisite for https://github.com/LycheeOrg/Lychee-front/pull/318 and https://github.com/LycheeOrg/Lychee/pull/1494

 * **The upload dialog became more robust.**
    * No more inefficient and unreliable CSS selector to pick the _n_-th status entry to update. Instead, DOM-elements are now cached and are accessible by the key of the status message (i.e. the path of the file related to the message).
    * Scrolling to the current status entry became more reliable (for me it was always off by some pixels). Instead of calculating the position manually based on magic pixel distances as in https://github.com/LycheeOrg/Lychee-front/blob/ab0f314cac720b28fbe1689a482bdd3af0f18efe/scripts/main/upload.js#L212-L214 we simply use `row.listEntry.scrollIntoView(upload.SCROLL_OPTIONS);` which we can now do, because we work with proper DOM elements (here: `listEntry`).

   Note, `upload.js` is still a mess with a lot of code duplication for importing photos from server and uploading via web browser. This needs to be fixed in another PR, not this one.

 * **Simplified structure of modal dialogs and simplified, more consistent CSS**: (more about that below)

### Structure of modal dialogs and CSS

The current way how dialogs were constructed was ways too complicated. Previously, a single checkbox with a label to the left and an indented description below was constructed like this

```html
<div class='choice'>
	<label>
		<input type='checkbox' name='grants_full_photo'>
		<span class='checkbox'>${build.iconic("check")}</span>
		<span class='label'>${lychee.locale["XYZ"]}</span>
	</label>
	<p>${lychee.locale["XYZ"]}</p>
</div>
```

Note the redundant use of the class `'choice' ` in `div` and `span`. The unnecessary nesting and the additional `span` to build the icon for the checkbox. Now the same thing can be achieved via

```html
<div class='input-group compact-inverse'>
	<label>${lychee.locale["XYZ"]}</label>
	<input type='checkbox' name='requires_link' />
	<p>${lychee.locale["XYZ"]}</p>
</div>
```

Moreover, previously every dialog was more or less a unique specimen. I went back to the drawing board and considered all dialogs Lychee uses and tried to carve out a "common structure". The result and assumptions are documented in the CSS: https://github.com/LycheeOrg/Lychee-front/blob/09ce761768894ceb3efd2de0b7d370561b576bce/styles/main/_form.scss#L1-L14 and https://github.com/LycheeOrg/Lychee-front/blob/09ce761768894ceb3efd2de0b7d370561b576bce/styles/main/_form.scss#L190-L235

This new approach and the simplified HTML makes the layout of the dialogs more robust and actually allows to nest form elements with several levels of indention (in case this is every required).

Previously, the layout worked as follows:

Screenshot_20220915_200445

Screenshot_20220915_200633

The actual `<input>` element had been sitting between the `<label>` which visualized the checkbox and the label to the right. It was invisible but contributed 14px of horizontal spacing to the gap in between the visual representation of the checkbox and its label. The description below has been responsible for its own indentation by 35px on the left to visually align with the label above. Hence, the CSS created an unspoken interdependence between the with of the visual representation of the checkbox (16px), the the width of the invisible real checkbox (14px) and a 1px border of the label which in sum "accidentally" happen to equal the 35px of indention of the description.

The new approach looks like this

Screenshot_20220915_202221

It has become much easier as it lifts any implicit interdependence. Now the `div` which creates the "input group" is responsible for the correct indention which applies to all elements in the input group simultaneously. Also the `<input>` element is actually the input element.


scripts/3rd-party/basicModal.js
```diff
@@ -87,48 +174,140 @@
...
 
 /**
- * @callback ModalDialogButtonCB
- * @param {Object} [values] an associative object with the values of all HTML
- *                          input elements; see {@link basicModal.getValues}
+ * Returns `true`, if the Action button is hidden.
+ *
+ * Note, this method is not exactly the opposite of
+ * {@link basicModal#isActionButtonVisible}.
+ * This method only returns `true` if the dialog own an Action button which
```
kamil4, line 212:
```suggestion
 * This method only returns `true` if the dialog owns an Action button which
```

scripts/3rd-party/basicModal.js
```diff
@@ -87,48 +174,140 @@
...
+ * @memberOf basicModal
+ * @returns {boolean}
+ */
+
+/**
+ * Returns `true`, if the Cancel button is hidden.
+ *
+ * Note, this method is not exactly the opposite of
+ * {@link basicModal#isCancelButtonVisible}.
+ * This method only returns `true` if the dialog own a Cancel button which
```
kamil4, line 279:
```suggestion
 * This method only returns `true` if the dialog owns a Cancel button which
```

kamil4 (changes requested):
_3 / 27 files viewed_ -- I'll try to keep working on this tomorrow, but this may take a while to review, especially since I expect that I will want to try it out as well.

package.json
```diff
@@ -19,7 +19,7 @@
 		"node": ">= 10"
 	},
 	"dependencies": {
-		"basicmodal": "https://github.com/LycheeOrg/basicModal/archive/refs/tags/v3.3.9.tar.gz",
+		"basicmodal": "LycheeOrg/basicModal#d1185e7",
```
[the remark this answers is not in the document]
nagmat84: This isn't meant to be a permanent situation. This PR relies on some bug fixes to basicModal which are not yet merged into the `master` branch of basicModal but are part of the open PR https://github.com/LycheeOrg/basicModal/pull/5. Hence, I needed to point to the tip of https://github.com/LycheeOrg/basicModal/pull/5 during development of this PR as long as the upstream PR hasn't merged either.

The idea is as follows: Squash and merge https://github.com/LycheeOrg/basicModal/pull/5 into `master` of modalDialog first, make a release of modalDialog, then change the package file here and switch over to the new release of basicModal.

Basically the situation here is comparable to the situation between Lychee-front and Lychee.

scripts/3rd-party/basicModal.js
```diff
@@ -7,44 +7,130 @@
...
+ *                                                    via {@link close}
+ * @property {ModalDialogButtonsData} buttons         configuration data for the main action and
+ *                                                    cancel button
+ * @property {ModalDialogReadyCB}     [readyCB=null]  callback to be called after the dialog
+ *                                                    has become visible and ready for user input
+ */
+
+/**
+ * @callback ModalDialogReadyCB
+ * @param {ModelDialogFormElements} htmlElements a dictionary that maps names to form elements
```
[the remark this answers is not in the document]
nagmat84: Fixed by https://github.com/LycheeOrg/Lychee-front/pull/322/commits/f301e2dcb737083d2c83c63e9b03388be18de056.

scripts/3rd-party/basicModal.js
```diff
@@ -7,44 +7,130 @@
...
+ * @returns {void}
+ */
+
+/**
+ * @callback ModalDialogClosedCB
+ * @returns {void}
+ */
+
+/**
+ * @typedef ModelDialogFormElements
```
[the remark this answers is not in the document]
nagmat84: Fixed by https://github.com/LycheeOrg/Lychee-front/pull/322/commits/f301e2dcb737083d2c83c63e9b03388be18de056.

scripts/main/lychee.js
```diff
@@ -213,24 +226,45 @@ lychee.logs = function () {
...
-				<p>${sprintf(lychee.locale["ABOUT_DESCRIPTION"], lychee.website)}</p>
-			  `;
+	const aboutDialogBody = `
+		<h1>Lychee <span class="version-number"></span></h1>
+		<p class="update-status up-to-date"><a target='_blank' href='${lychee.updateURL}'></a></p>
+		<h2></h2>
+		<p class="about-desc"></p>`;
+
+	/**
+	 * @param {ModelDialogFormElements} formElements
```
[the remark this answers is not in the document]
nagmat84: Fixed by https://github.com/LycheeOrg/Lychee-front/pull/322/commits/f301e2dcb737083d2c83c63e9b03388be18de056.

scripts/main/lychee.js
```diff
@@ -425,55 +467,69 @@ lychee.login = function (data) {
...
+				<input class='text' name='password' autocomplete='current-password' type='password' data-tabindex='${tabindex.get_next_tab_index()}'>
+			</div>
+		</form>
+		<p class='version'>Lychee <span class='version-number'></span><span class="update-status up-to-date"> &#8211; <a target='_blank' href='${
+			lychee.updateURL
+		}' data-tabindex='-1'></a></span></p>
+		`;
+
+	/**
+	 * @param {ModelDialogFormElements} formElements
```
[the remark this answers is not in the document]
nagmat84: Fixed by https://github.com/LycheeOrg/Lychee-front/pull/322/commits/f301e2dcb737083d2c83c63e9b03388be18de056.

scripts/main/password.js
```diff
@@ -49,24 +43,30 @@ password.getDialog = function (albumID, callback) {
...
-			  </p>
-			  `;
+	const enterPasswordDialogBody = `
+		  <p></p>
+		  <form>
+		  	<div class="input-group stacked"><input name='password' class='text' type='password'></div>
+		  </form>`;
+
+	/**
+	 * @param {ModelDialogFormElements} formElements
```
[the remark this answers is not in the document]
nagmat84: Fixed by https://github.com/LycheeOrg/Lychee-front/pull/322/commits/f301e2dcb737083d2c83c63e9b03388be18de056.

scripts/main/settings.js
```diff
@@ -493,77 +497,122 @@ settings.save_enter = function (e) {
...
-		$("#button_disable_token").on(lychee.getEventName(), function () {
-			$("#apiToken").text(lychee.locale["TOKEN_WAIT"]);
-			api.post("User::unsetToken", {}, function () {
-				token = "";
-				lychee.user.has_token = false;
-				updateTokenDialog();
-			});
-		});
+	/**
+	 * @param {ModelDialogFormElements} formElements
```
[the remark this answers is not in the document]
nagmat84: Fixed by https://github.com/LycheeOrg/Lychee-front/pull/322/commits/f301e2dcb737083d2c83c63e9b03388be18de056.

scripts/main/lychee.js
```diff
@@ -213,24 +226,45 @@ lychee.logs = function () {
...
 lychee.aboutDialog = function () {
-	const msg = lychee.html`
-				<h1>Lychee ${lychee.version}</h1>
-				<div class='version'><span><a target='_blank' href='${lychee.updateURL}'>${lychee.locale["UPDATE_AVAILABLE"]}</a></span></div>
-				<h1>${lychee.locale["ABOUT_SUBTITLE"]}</h1>
-				<p>${sprintf(lychee.locale["ABOUT_DESCRIPTION"], lychee.website)}</p>
-			  `;
+	const aboutDialogBody = `
+		<h1>Lychee <span class="version-number"></span></h1>
+		<p class="update-status up-to-date"><a target='_blank' href='${lychee.updateURL}'></a></p>
```
kamil4, line 231: Do we need this `up-to-date` here, considering that we programmatically add it as-needed in `initAboutDialog` below?

scripts/main/lychee.js
```diff
@@ -425,55 +467,69 @@ lychee.login = function (data) {
...
+		<a id='signInKeyLess' class="button"><svg class='iconic'><use xlink:href='#key'/></svg></a>
+		<form class="force-first-child">
+			<div class="input-group stacked">
+				<input class='text' name='username' autocomplete='username' type='text' autocapitalize='off' data-tabindex='${tabindex.get_next_tab_index()}'>
+			</div>
+			<div class="input-group stacked">
+				<input class='text' name='password' autocomplete='current-password' type='password' data-tabindex='${tabindex.get_next_tab_index()}'>
+			</div>
+		</form>
+		<p class='version'>Lychee <span class='version-number'></span><span class="update-status up-to-date"> &#8211; <a target='_blank' href='${
```
kamil4, line 480: Same question about the need for `up-to-date` here as before.

kamil4 (commented):
Making progress (11 / 27 files viewed) but still some way to go...

scripts/main/album.js
```diff
@@ -771,35 +838,62 @@ album.setSorting = function (albumID) {
...
+			</div>
+		</form>`;
+
+	/**
+	 * @param {ModalDialogFormElements} formElements
+	 * @param {HTMLDivElement} dialog
+	 * @returns {void}
+	 */
+	const initSetAlbumSortingDialog = function (formElements, dialog) {
+		formElements.sorting_col.labels[0].textContent = "Attribute";
```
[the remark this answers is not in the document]
nagmat84: In theory yes. When I have been working on this PR, I recognized that probably 20% of all strings which should be localized are not yet enabled for localization. This one is not the only string. There are also still more labels, tooltips, etc.

This is not a regression.

I did not feel like fixing that problem within this PR, too. Because, I know myself. If I had started on that, then I would have gone the whole way which would have increased and extended this PR extremely. Hence, I decided to stick closely to the topic of this PR and to not touch anything else which crosses my path in between.

scripts/main/album.js
```diff
@@ -771,35 +838,62 @@ album.setSorting = function (albumID) {
...
+		formElements.sorting_col.item(1).textContent = lychee.locale["SORT_PHOTO_SELECT_1"];
+		formElements.sorting_col.item(2).textContent = lychee.locale["SORT_PHOTO_SELECT_2"];
+		formElements.sorting_col.item(3).textContent = lychee.locale["SORT_PHOTO_SELECT_3"];
+		formElements.sorting_col.item(4).textContent = lychee.locale["SORT_PHOTO_SELECT_4"];
+		formElements.sorting_col.item(5).textContent = lychee.locale["SORT_PHOTO_SELECT_5"];
+		formElements.sorting_col.item(6).textContent = lychee.locale["SORT_PHOTO_SELECT_6"];
+		formElements.sorting_col.item(7).textContent = lychee.locale["SORT_PHOTO_SELECT_7"];
+
+		formElements.sorting_order.labels[0].textContent = "Order";
```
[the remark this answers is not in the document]
nagmat84: See the previous answer.

scripts/main/lychee.js
nagmat84: I added it, to theoretically avoid flicker in short time between the dialog has been opened and before the method `initAboutDialog` has been completely executed. If the  class `up-to_date` is not defined on the paragraph, the paraph is initially visible and then vanished if the class is added later programmatically. However, I say "theoretically", because I never experienced it in practice. Probably, modern systems are simply fast enough.

Yes, you a right, in the code

```php
if (lychee.update_available) {
	updClassList.remove("up-to-date");
} else {
	updClassList.add("up-to-date");
}
```

below, I could remove the `else`-branch, because `"up-to-date"` there is no point in adding it. Shall I?

An an optimal world, I world we would change `basicModal` as follows: first construct the the dialog on the shadow DOM (we do that), then call the provided initializing callback and when the callback has completed mount the shadow DOM into the actual DOM. This would avoid any flicker under all circumstances. (Also for more complex dialogs.) It would ensure that the whole dialog becomes only visible at once after it has been fully initialized. May be I do that.

scripts/main/lychee.js
nagmat84: See previous answer.

scripts/main/photo.js
```diff
@@ -1007,109 +1016,92 @@ photo.setLicense = function (photoID) {
...
+	 * @param {ModalDialogFormElements} formElements
+	 * @param {HTMLDivElement} dialog
+	 */
+	const initGetPhotoArchiveDialog = function (formElements, dialog) {
+		Object.entries(kind2VariantAndLocalizedLabel).forEach(function ([kind, [variant, lLabel]]) {
+			/** @type {HTMLAnchorElement} */
+			const button = dialog.querySelector('a[data-photo-kind="' + kind + '"]');
+			/** @type {?SizeVariant} */
+			const sv = myPhoto.size_variants[variant];
+			if (!!sv) {
```
kamil4, line 1074: Any particular reason for using `!!` here?

scripts/main/photo.js
```diff
@@ -1183,85 +1172,85 @@ photo.showDirectLinks = function (photoID) {
...
+		"</form>";
+
+	/**
+	 * @param {ModalDialogFormElements} formElements
+	 * @param {HTMLDivElement} dialog
+	 */
+	const initShowDirectLinksDialog = function (formElements, dialog) {
+		formElements.view.value = photo.getViewLink(photoID);
+		formElements.view.previousElementSibling.textContent = lychee.locale["PHOTO_VIEW"];
+		formElements.view.nextElementSibling.title = "Copy to clipboard";
```
kamil4, line 1212: This string should be localized (and I think it was in the replaced code?).

scripts/main/photo.js
```diff
@@ -1183,85 +1172,85 @@ photo.showDirectLinks = function (photoID) {
...
+		dialog.querySelector("p").textContent = lychee.locale["PHOTO_DIRECT_LINKS_TO_IMAGES"];
+
+		for (const type in localizations) {
+			/** @type {?SizeVariant} */
+			const sv = photo.json.size_variants[type];
+			if (sv !== null) {
+				formElements[type].value = lychee.getBaseUrl() + sv.url;
+				formElements[type].previousElementSibling.textContent = localizations[type] + " (" + sv.width + "×" + sv.height + ")";
+				formElements[type].nextElementSibling.title = "Copy to clipboard";
```
kamil4, line 1221: Same here (localization).

scripts/main/photo.js
```diff
@@ -1183,85 +1172,85 @@ photo.showDirectLinks = function (photoID) {
...
+		}
 
-	msg += lychee.html`
-		</div>
-		</div>
-	`;
+		if (photo.json.live_photo_url !== null) {
+			formElements.live.value = lychee.getBaseUrl() + photo.json.live_photo_url;
+			formElements.live.previousElementSibling.textContent = lychee.locale["PHOTO_LIVE_VIDEO"];
+			formElements.live.nextElementSibling.title = "Copy to clipboard";
```
kamil4, line 1234: Same here (localization).

scripts/main/photo.js
```diff
@@ -564,173 +559,151 @@ photo.setStar = function (photoIDs, isStarred) {
...
+		} else {
+			// Private album
+			dialog.querySelector("p#ppp_dialog_no_edit_expl").remove();
+			dialog.querySelector("p#ppp_dialog_global_expl").textContent = lychee.locale["PHOTO_EDIT_GLOBAL_SHARING_TEXT"];
+			// Initialize values of detailed settings according to global
+			// configuration.
+			formElements.is_public.checked = photo.json.is_public !== 0;
+			formElements.grants_full_photo.checked = lychee.full_photo;
+			formElements.requires_link.checked = lychee.public_photos_hidden;
+			formElements.is_downloadable.checked = !!album.downloadable;
```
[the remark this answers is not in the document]
nagmat84: Yes and no.

Bo, `lychee.downloadable` is also wrong, because it online affects public photos in an private album.

Yes, `album.downloadable` is only meaningful for public albums. The idea here is that the dialog shows the settings which would apply, if the album was public. The fact that `album.downloadable` has no effect is visualized by the control elements being greyed out.

scripts/main/album.js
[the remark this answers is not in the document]
kamil4: I agree that you should not be fixing existing localization issues in this PR.

However, this particular case very much _does_ look like a regression to me, which is why I brought it up. In the code you replaced, there was `lychee.locale["SORT_PHOTO_BY"]`; you replaced it with hardcoded `"Attribute"` and `"Order"` strings...

scripts/main/photo.js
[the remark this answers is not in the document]
kamil4: > Bo, lychee.downloadable is also wrong, because it online affects public photos in an private album.

I'm not sure if I understood you correctly given the typos, but this _is_ a case of a private album, hence we should be consulting `lychee.downloadable`.

> The idea here is that the dialog shows the settings which would apply, if the album was public.

No. For private albums, the dialog is supposed to show the global settings (`lychee.*`), since they are the ones applicable to public photos in private albums. album-level settings are irrelevant in that case and would only add to confusion.

I think I'm the one who coded this dialog in the first place, so let me summarize how I intended it to behave:
* If the album is public, we mostly mirror the album settings here, but everything is greyed out/disabled. There is a message at the top that explains why none of the settings can be changed.
* If the album is private but the photo is public, we mostly mirror the global settings here, but everything is again greyed out (apart from the "public" switch at the top) and there's a message in the middle that explains why.
* If the album is private *and* the photo is private, the (still greyed out) checkboxes are empty, since no settings apply in that case (the photo is private). From reading the code, your code does not handle that case though and it will fill in the checkboxes with global settings. I'm not sure if I like that but I reserve the judgement until I've had a chance to actually try it out.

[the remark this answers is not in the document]
kamil4: On thinking about it a little more, that last case above (of private album _and_ private photo) perhaps _should_ show global settings (as it does in your case), since those settings _will_ be applicable if the album is shared with another user...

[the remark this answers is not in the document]
nagmat84: > > No, `lychee.downloadable` is also wrong, because it online affects public photos in an private album.
> 
> I'm not sure if I understood you correctly given the typos, but this _is_ a case of a private album, hence we should be consulting `lychee.downloadable`.

Yes, it is a private _album_. The point is, the album dialog shows the _album_ settings, not the settings which apply to a public photo in that private album. If users want to see the settings of a photo, the need to open the photo and look at the photo settings of that.

> > The idea here is that the dialog shows the settings which would apply, if the album was public.
> 
> No. For private albums, the dialog is supposed to show the global settings (`lychee.*`), since they are the ones applicable to public photos in private albums. album-level settings are irrelevant in that case and would only add to confusion.

I guess this is the point were we are on different pages. I believe the album dialog should always show the settings of the album, even if some of the settings are currently irrelevant which is visually indicated by greyed out controls. Personally, I would deem it confusing, if the album dialog happens to show (global) photo settings only because some album settings are not applicable at the moment.
 
> I think I'm the one who coded this dialog in the first place, so let me summarize how I intended it to behave:
> 
>     * If the album is public, we mostly mirror the album settings here, but everything is greyed out/disabled. There is a message at the top that explains why none of the settings can be changed.

OK, and here is the point where I notice that you are speaking about the "photo dialog" and not the "album dialog" as I could have noticed by looking at the file name `photos.js`. I am really worn out.

scripts/main/settings.js
```diff
@@ -19,10 +19,14 @@ settings.createLogin = function () {
...
-		let htmlBody = "<p>" + lychee.locale["ERROR_LOGIN"] + "</p>";
-		htmlBody += lycheeException ? "<p>" + lycheeException.message + "</p>" : "";
 		basicModal.show({
-			body: htmlBody,
+			body: "<p></p><p></p>",
+			readyCB: (formElement, dialog) => {
+				/** @type {NodeList<HTMLParagraphElement>} */
+				const paragraphs = dialog.querySelectorAll("p");
+				paragraphs.item(0).textContent = lychee.locale["ERROR_LOGIN"];
+				paragraphs.item(1).textContent = lycheeException ? "<p>" + lycheeException.message + "</p>" : "";
```
kamil4, line 28:
```suggestion
				paragraphs.item(1).textContent = lycheeException ? lycheeException.message : "";
```

`body` already contains a second set of `<p>`/`</p>`, doesn't it?

scripts/main/settings.js
```diff
@@ -457,32 +468,25 @@ settings.save_enter = function (e) {
...
-	action.msg = lychee.html`<p style="color: #d92c34; font-size: 1.3em; font-weight: bold; text-transform: capitalize; text-align: center;">${lychee.locale["SAVE_RISK"]}</p>`;
-
-	cancel.title = lychee.locale["CANCEL"];
-
-	action.fn = function () {
-		settings.save(settings.getValues("#fullSettings"));
-		basicModal.close();
-	};
+	const saveSettingsConfirmationDialogBody =
+		'<p style="color: #d92c34; font-size: 1.3em; font-weight: bold; text-transform: capitalize; text-align: center;"></p>';
```
kamil4, line 472:
```suggestion
		// TODO: move the style to the style file, where it belongs.
		'<p style="color: #d92c34; font-size: 1.3em; font-weight: bold; text-transform: capitalize; text-align: center;"></p>';
```

I know it was there before, but it's just too ugly to silently accept...

scripts/main/album.js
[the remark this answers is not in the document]
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/66619a4ba4a6c5039fb40a3cd8c4d7e627a00fdc.

scripts/main/album.js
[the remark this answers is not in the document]
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/66619a4ba4a6c5039fb40a3cd8c4d7e627a00fdc.

scripts/main/lychee.js
nagmat84: I removed the unnecessary `else`-branch in https://github.com/LycheeOrg/Lychee-front/commit/8279248d69de6ec8e4a7df7f597cf302c25e3cb2.

scripts/main/lychee.js
nagmat84: I removed the unnecessary `else`-branch in https://github.com/LycheeOrg/Lychee-front/commit/8279248d69de6ec8e4a7df7f597cf302c25e3cb2.

scripts/main/album.js
```diff
@@ -868,111 +961,150 @@ album.setProtectionPolicy = function (albumID) {
...
-			$('.basicModal .choice input[name="grants_full_photo"]').prop("checked", lychee.grants_full_photo);
-			$('.basicModal .choice input[name="requires_link"]').prop("checked", false);
-			$('.basicModal .choice input[name="is_downloadable"]').prop("checked", lychee.is_downloadable);
-			$('.basicModal .choice input[name="is_share_button_visible"]').prop("checked", lychee.is_share_button_visible);
-			$('.basicModal .choice input[name="has_password"]').prop("checked", false);
-			$('.basicModal .choice input[name="passwordtext"]').hide();
+			formElements.grants_full_photo.checked = lychee.grants_full_photo;
+			formElements.requires_link.checked = false;
+			formElements.is_downloadable.checked = lychee.is_downloadable;
+			formElements.is_share_button_visible.checked = lychee.is_share_button_visible;
```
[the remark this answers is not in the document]
nagmat84: > I wish we were getting a warning about such cases from some code checker though...

That's future work. Something like this is possible. A proper IDE and CI tools would be able to catch those things, but only if we would use proper JS modules and `use "strict";`. Without that the code above is actually not wrong. I am aiming into that direction, but it will still take some time.

scripts/main/photo.js
nagmat84: No, except for my laziness to check with the specs what happens if `sv` is not `null` but `undefined`. But I guess not using `!!` is fine, too. Shall I change it?

scripts/main/photo.js
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/commit/7065ddba2d9db477f9940c9ce6a875fd4e477840.

scripts/main/photo.js
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/commit/7065ddba2d9db477f9940c9ce6a875fd4e477840.

scripts/main/photo.js
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/commit/7065ddba2d9db477f9940c9ce6a875fd4e477840.

scripts/main/settings.js
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/commit/b60fdedf62f76fa77c29ff7dda2e3ce08e9d79e6.

Your are totally right. It is even harmful, because `<p>` would be shown literally. This is how `textContent` works.

scripts/main/settings.js
nagmat84: Agreed.

I am even not sure why we want to increase the font size and make it bold. That is too much for my taste. IMHO, the "normal" warning dialog should be sufficient.

scripts/main/photo.js
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/d4bdecdb2c64b19e3979cc65189fc21d8aa7e829.

scripts/main/upload.js
```diff
@@ -571,44 +732,47 @@ upload.start = {
...
-								// If we do not have a current row yet, we
-								// simply append it to the list of rows
-								// (this might happen if the event occurs
-								// before the first progress report)
-								jqEventRow = $(build.uploadNewFile(report.path || "General")).appendTo(jqRows);
-								topSkip += jqEventRow.outerHeight();
+								// The event report does not refer to a
+								// specific directory.
+								// We insert the event row _before_ the row
+								// before the last row, so that the latest
```
kamil4, line 770:
```suggestion
								// We insert the event row _before_ the last
								// row, so that the latest
```

styles/main/_form.scss
```diff
@@ -0,0 +1,373 @@
+// A HTML `<form>` is expected to consists of a sequence of one ore more
```
kamil4, line 1:
```suggestion
// A HTML `<form>` is expected to consist of a sequence of one or more
```

styles/main/_form.scss
```diff
@@ -0,0 +1,373 @@
...
+// 1. Form elements (general styling, independent of input group)
+
+input,
+div.select,
+select,
+textarea,
+output {
+	display: inline-block;
+	/* we must position this element so that it becomes the nearest
+   * positioned ancestor for ::before, ::after and :checked::before */
```
kamil4, line 23:
```suggestion
	/* we must position this element so that it becomes the nearest
	 * positioned ancestor for ::before, ::after and :checked::before */
```

styles/main/_form.scss
```diff
@@ -0,0 +1,373 @@
...
+	position: relative;
+}
+
+div.select > select {
+	// The element `<select>` is hard to style with CSS, because most browser
```
kamil4, line 28:
```suggestion
	// The element `<select>` is hard to style with CSS, because most browsers
```

styles/main/_form.scss
```diff
@@ -0,0 +1,373 @@
...
+select option {
+	color: $colorFormElementFg;
+	background-color: $colorFormElementBg;
+	// input elements always have zero margin:
+	// - if they are the first/last child, they need a zero-margin
+	//   anyway
+	// - if they are preceded by a label or followed by a paragraph,
+	//   then the label has a bottom or the paragraph a top margin
+	margin: 0;
+	// Form elements are always special, because browser are very
```
kamil4, line 53:
```suggestion
	// Form elements are always special, because browsers are very
```

styles/main/_form.scss
```diff
@@ -0,0 +1,373 @@
...
+//  - zero or more optional `<p>` elements with a descriptive text
+//  - zero or more optional `<div>` with nested input groups
+//
+// One note about `<select>`: This element is hard to style with CSS,
+// because it is treated as a replaced element. Hence, `<select>` needs to
+// be wrapped into an additional `<div>` and most styles are applied to
+// this `<div>`.
+// See: https://developer.mozilla.org/en-US/docs/Learn/Forms/Advanced_form_styling
+//
+// An input group can have one out of four layouts which determines
```
kamil4, line 212:
```suggestion
// An input group can have one out of four layouts which deteriine
```

styles/main/_form.scss
```diff
@@ -0,0 +1,373 @@
...
+//
+//  2. `compact`
+//      The label and the form element are on the same horizontal line.
+//      The other elements follow below with a hanging indention and
+//      are aligned to the form element
+//
+//  3. `compact-inverse`
+//      The form element and the label are on the same horizontal line.
+//      The other elements follow below with a hanging indention and are
+//      aligned to the form element
```
kamil4, line 227: The explanation for (2) and (3) is currently the same, which is unhelpful.

Also, I'm a bit surprised by the word _indention_. While apparently an existing word, I think I've always seen _indentation_ used in this context?

styles/main/_form.scss
```diff
@@ -0,0 +1,373 @@
...
+			display: block;
+			// if a label exists, then it is always the first child,
+			// so we only need a bottom margin, no top margin
+			margin-bottom: 6px;
+		}
+
+		> label > input[type="text"],
+		> label > input[type="password"] {
+			// in stacked layout, text input needs a little bit more space,
+			// if preceded by a label, otherwise it looks to crowded
```
kamil4, line 286:
```suggestion
			// if preceded by a label, otherwise it looks too crowded
```

styles/main/_form.scss
```diff
@@ -0,0 +1,373 @@
...
+		> label {
+			display: block;
+			position: absolute;
+			margin: 0;
+			left: 0;
+			width: 108px; // 120px of padding - 12px normal "gap"
+			height: auto;
+			top: 3px; // adjustment for padding of text input and select
+			bottom: 0; // height should take the whole left padding
+			overflow-y: hidden; // truncate over-sized labels so that the do not spill into other areas
```
kamil4, line 313:
```suggestion
			overflow-y: hidden; // truncate over-sized labels so that they do not spill into other areas
```

kamil4 (commented):
I finally finished going over this PR. I would still like to see the results in a web browser before final approval though.

styles/main/_form.scss
d7415: I think "indentation" is more correct. "Indent" is more common because that's what Word uses. Also, e.g. `text-indent` in CSS.

nagmat84: > The explanation for (2) and (3) is currently the same, which is unhelpful.

Actually not :laughing: 

> 2. The **label** and the **form element** are on the same horizontal line.

vs.

> 3. The **form element** and the **label** are on the same horizontal line.

The order of words was swapped on purpose. But I agree that this was extremely subtle. :laughing: I made it more explicit in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/4320beb8e5c2347321aa019d4561b012de73d9d7.

> Also, I'm a bit surprised by the word indention. While apparently an existing word, I think I've always seen indentation used in this context?

Also fixed in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/4320beb8e5c2347321aa019d4561b012de73d9d7.

scripts/main/upload.js
```diff
@@ -486,27 +591,93 @@ upload.start = {
...
 
-		const importDialogSetupCB = function () {
-			const $delete = $(choiceDeleteSelector);
-			const $symlinks = $(choiceSymlinkSelector);
-			const $duplicates = $(choiceDuplicateSelector);
-			const $resync = $(choiceResyncSelector);
+		/**
+		 * @typedef ImportFromServerDialogFormElements
+		 *
+		 * @property {HTMLInputElement} path
```
kamil4, line 597:
```suggestion
		 * @property {HTMLInputElement} paths
```

scripts/main/upload.js
```diff
@@ -486,27 +591,93 @@ upload.start = {
...
 
+		/**
+		 * @param {ImportFromServerDialogFormElements} formElements
+		 * @param {HTMLDivElement} dialog
+		 * @returns {void}
+		 */
+		const initImportFromServerDialog = function (formElements, dialog) {
+			dialog.querySelector("p").textContent = lychee.locale["UPLOAD_IMPORT_SERVER_INSTR"];
+			formElements.path.placeholder = lychee.locale["UPLOAD_ABSOLUTE_PATH"];
+			formElements.path.value = lychee.location + "uploads/import/";
```
kamil4, line 612:
```suggestion
			formElements.paths.placeholder = lychee.locale["UPLOAD_ABSOLUTE_PATH"];
			formElements.paths.value = lychee.location + "uploads/import/";
```

scripts/main/upload.js
```diff
@@ -751,63 +915,62 @@ upload.start = {
...
-					${lychee.locale["UPLOAD_IMPORT_RESYNC_METADATA_EXPL"]}
-				</p>
-			</div>
-		`;
+			upload.showProgressReportDialog(lychee.locale["UPLOAD_IMPORT_SERVER"], [], runUpload, cancelUpload);
+		}; // importFromServer
+
+		/** @param {ServerImportDialogResult} data */
+		const processImportFromServerDialog = function (data) {
+			if (!data.path.trim()) {
```
kamil4, line 923:
```suggestion
			if (!data.paths.trim()) {
```

kamil4 (changes requested):
"Import from Server" appears to be untested as it fails in multiple ways. I started fixing it (see my suggestions) but I just seem to uncover more bugs... In particular, method `action` (under `server`) needs to be renamed to `importFromServer`, but then `data.paths.trim()` fails because `data.paths` is an array, not a string. At that point I gave up as it's getting late here...

styles/main/_form.scss
kamil4: Right, a typo in a suggestion :flushed:.

nagmat84:
> "Import from Server" appears to be untested as it fails in multiple ways. I started fixing it (see my suggestions) but I just seem to uncover more bugs... In particular, method `action` (under `server`) needs to be renamed to `importFromServer`, but then `data.paths.trim()` fails because `data.paths` is an array, not a string. At that point I gave up as it's getting late here...

Thanks for being thorough. It is not your job to fix it, but mine. I'll take it from here.

I assume I know what happened. At some point while working on this PR, the PR https://github.com/LycheeOrg/Lychee-front/pull/308 got merged into master which introduced the option to use multiple, space separated paths. This PR introduced a regression on master, because it did not properly deal with paths which contained a space in themselves. I noticed that the (normal) upload did not work as expected. I fixed this regression in https://github.com/LycheeOrg/Lychee-front/commit/3e7a46955d8f9edf925bcd4aeaa3070d92b6695d.

I thought that I had merged both PRs into this one properly and I also believed I had tested upload and import after that, but obviously I only checked the upload. Shame on me.

styles/main/_form.scss
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/9e941bbfbc26240a8b3e6bfa63e7546fe59123be.

nagmat84:
> "Import from Server" appears to be untested as it fails in multiple ways. I started fixing it (see my suggestions) but I just seem to uncover more bugs... In particular, method `action` (under `server`) needs to be renamed to `importFromServer`, but then `data.paths.trim()` fails because `data.paths` is an array, not a string. At that point I gave up as it's getting late here...

I fixed it in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/f181f2c8b8739f76108b2fa6e96ce82880a4db1b.

It was indeed a failed merge. I don't know how that happened. Luckily I could restore the correct code from an older commit. @kamil4 You already found 95% of the problems yourself in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/a9f5ba1823cac92ec34a12cb9871542dea5fe3be.

> In particular, method `action` (under `server`) needs to be renamed to `importFromServer`, but then `data.paths.trim()` fails because `data.paths` is an array, not a string.

The code that was there was completely outdated. It suddenly also used jQuery selectors again which actually was an alarm sign to me that something really, really bad had happened. The remaining bug fix is in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/f181f2c8b8739f76108b2fa6e96ce82880a4db1b. It was mostly about deleting old code which had sneaked in again.

It already had been correct in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/ef3eddcfdc2c70aec791b83652f3795ae4150298 once. I hope this is the only "accident" which happened.

package.json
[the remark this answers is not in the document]
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/88fa3e66e7fb7fec2edc1aa2d4900d9520486f02. Now it is
```json
"basicmodal": "LycheeOrg/basicModal#v4.0.0"
```
after https://github.com/LycheeOrg/basicModal/pull/8 has been merged and after I have created a new version tag after more than 4.5 years. :sunglasses:

nagmat84:
@kamil4 I solved two of the three issues which you raised [on Gitter at Oct 15th 2022 07:17](https://gitter.im/LycheeOrg/Developers?at=634a426af00b697fec4141bd).

 - Wrt. to auto-focus: Fixed by https://github.com/LycheeOrg/basicModal/pull/10. This was actually an oversight in basicModal. The commit https://github.com/LycheeOrg/Lychee-front/pull/322/commits/1211769e3d60ea234101144cf094fd28491c6769 here is only included to illustrate the the fix for basicModal works until that has been merged upstream
 - Wrt. to search box: Fixed by https://github.com/LycheeOrg/Lychee-front/pull/322/commits/3b75276857082d6e8a23e6620716a7b005ed3d91. This was a problem with the specificity of CSS ruies.

kamil4:
Somehow the commit https://github.com/LycheeOrg/Lychee-front/commit/1211769e3d60ea234101144cf094fd28491c6769 has no effect for me; even if I delete the whole `node_modules` and refetch everything using `npm install`, it still fetches the old version. I had to copy the files over manually. Now the focusing indeed works.

I see an undesirable change in the look in the individual photo download dialog.

Old:
old
New:
new

As you can see, the black bar on top of each button is missing. Another, more subtle change is that whereas before the buttons were subtly darker than the background, now they fully blend in. From a quick look, on `master` the buttons are of the more generic `basicModal__button` class, whereas on your branch they are `button` class. There's also another, though probably unrelated regression, in that the space in front of the opening parenthesis is gone.

kamil4:
When importing from the server, any errors (e.g., if there's an `index.html` file in the folder) are now reported _underneath_ the "Finished" line for the whole folder where they occurred, whereas before they would end up above. I believe I originally coded it that way with the intention that the output be chronological in nature from top to bottom, but honestly I don't have a strong preference anymore. I'm pointing it out just in case somebody else had a strong opinion...

nagmat84:
> Somehow the commit [1211769](https://github.com/LycheeOrg/Lychee-front/commit/1211769e3d60ea234101144cf094fd28491c6769) has no effect for me; even if I delete the whole `node_modules` and refetch everything using `npm install`, it still fetches the old version.

That is somehow expected. I noticed the same problem. You must run `npm update` instead of `npm install`. The difference between both commands for JS corresponds to the difference between `composer udpate` and `composer install` for PHP. `npm install` only reads `package-lock.json` and instalsl the packages listed in there with their exact locked version. `npm update` reads `package.json`, resolves the dependency tree, writes a new `package-lock.json` for later use by `npm install` and then installs the packages.

However, for some reason we don't commit `package-lock.json`, but only `package.json`. Hence, `npm install` won't work as expected because it will always install the (potentially outdated) package version you have installed locally. I don't know why we do it that way. It is inconsistent to our approach for composer where we also commit the "locked" version. This was introduced by @ildyria in https://github.com/LycheeOrg/Lychee-front/commit/8b6a10ee8554846b9f9d25e5fbb1b354d68f7a54.

ildyria:
> This was introduced by @ildyria in https://github.com/LycheeOrg/Lychee-front/commit/8b6a10ee8554846b9f9d25e5fbb1b354d68f7a54.

I don't remember why I did this. :see_no_evil:

nagmat84:
> I see an undesirable change in the look in the individual photo download dialog.
> 
> [...]
>
> As you can see, the black bar on top of each button is missing. [...] From a quick look, on `master` the buttons are of the more generic `basicModal__button` class, whereas on your branch they are `button` class.

The different style classes are not the cause of the regression. They have been there before and still are. The regression was rather a victim of my attempt to streamline the CSS. Do you remember the guy who was whining about the CSS being too complicated to create a custom color scheme? Same is true for everything else in this CSS.

Previously, there has been two different kind of settings at two different places which played together to achieve the former visual effect:

```CSS
.downloads .basicModal__button {
	box-shadow: inset 1px 0 0 rgba(0,0,0,.2);
	border-radius: 5px;
}

.basicModal__button {
	border-top: 1px solid rgba(0,0,0,.2);
}
```

The first CSS rule adds a "shadow" to the left side of the box. But as the third and fourth parameter ("blur radius" and "spread radius") are both zero, the "shadow" degenerates into something which creates the same visual effect as a left border of 1px width and a 20% transparent black. The same effect could have been achieved more explicitly (and efficiently computation wise) by `border-left: 1px solid rgba(0,0,0,.2)`. The second CSS rule adds a top border. Hence, both combined together give the appearance of a border at the left and at the top.

When I came across the second rule I just deleted it, because frankly it doesn't make much sense to have a border around a button only at the top.

This is a general problem with that whole CSS: It is a bunch of statements which are wildly scattered across several different rules but logically belong together. Even worse they depend on each other: if you change the color in one of the rules you need to remember to change to other rule, too, or the color of the border-like left shadow and the real top border won't match.  It's not separation of concerns but diffusion of concerns. :grin: 

Anyway I fixed it in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/e72d7a59a9619b9f522a6c8543754ba460868a0e but in a slightly different way than before. It is now simply:

```CSS
.button {
	border: 1px solid rgba(0,0,0,.2);
	border-radius: 5px;
}
```

So now we have a border everywhere (top, left, bottom, right).  @kamil4 Let me know if you like this one.

Honestly, I would like to throw out all the `shadow` statements from the CSS which do not create a real shadow but are just misused to create the visual appearance of a border. These are probably 90% of all shadow statements. But that's another PR.

> Another, more subtle change is that whereas before the buttons were subtly darker than the background, now they fully blend in.  

I cannot confirm that the background of the button was subtly darker than the background on `master`. My color picker also shows the same values. Are you sure that this is not only a visual illusion due to the darker border? But anyway, I added a rule to make the button actually a little bit darker. Now it is

```CSS
.button {
	border: 1px solid rgba(0,0,0,.2);
	border-radius: 5px;
	background-color: black(0.05);
}
```

> There's also another, though probably unrelated regression, in that the space in front of the opening parenthesis is gone.

Fixed in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/30fe8df858df58e27b3790b67af83c9b4ea17267.

kamil4:
I'll prefix this by saying that I'm not a graphic designer so I have no inherent understanding why things are the way they are or how they should be.

BUT (you knew that was coming, didn't you? :wink:).

> When I came across the second rule I just deleted it, because frankly it doesn't make much sense to have a border around a button only at the top.

I'm in agreement with you on this, I think. I've always found it strange how many buttons in Lychee look "unfinished", with no border at all on the bottom or the right. I'm all for doing something about that. And that applies not just to the download" buttons, but also, e.g., to "Copy to clipboard" in the photo's "direct links" dialog, or to the "key" button in the login dialog:

new3
new4


> So now we have a border everywhere (top, left, bottom, right). @kamil4 Let me know if you like this one.

I like that there's a border. I don't like that the buttons are completely "flat" now. That may be the current fashion, but I've actually always liked Lychee's subtle 3D look, I believe achieved with horizontal darker lines and brighter lines right next to each other. I believe on `master` this is achieved with the following two lines?

https://github.com/LycheeOrg/Lychee-front/blob/6471e787f242007900b7952d3990a5795974abeb/styles/main/_message.scss#L84-L85

Your change removes not only the dark shadow on the left (as you point out, it's unnecessary) but also the light one on top.

> Honestly, I would like to throw out all the shadow statements from the CSS which do not create a real shadow but are just misused to create the visual appearance of a border. These are probably 90% of all shadow statements. But that's another PR.

As I said, my personal preference is to preserve the current look (i.e., not go completely "flat"). If that can be achieved without shadows, that's fine with me, but it's not immediately obvious to me how to do that.

> I cannot confirm that the background of the button was subtly darker than the background on master. My color picker also shows the same values. Are you sure that this is not only a visual illusion due to the darker border?

Pretty sure; I just verified with the color picker in GIMP in my "old" image from my earlier post in this thread. I'm seeing a difference by a single shade (e.g., #414141 vs #424242), but it's there.

> But anyway, I added a rule to make the button actually a little bit darker.

I'm honestly not sure about that. I think a full border serves the same purpose better. I'm for eliminating that color difference and keeping the border instead.

> This is a general problem with that whole CSS: It is a bunch of statements which are wildly scattered across several different rules but logically belong together. Even worse they depend on each other: if you change the color in one of the rules you need to remember to change to other rule, too, or the color of the border-like left shadow and the real top border won't match. It's not separation of concerns but diffusion of concerns. :grin:

I would call the state of Lychee CSS consistent with the state of the rest of the front end: it is spaghetti code.

> You must run `npm update` instead of `npm install`

I try to avoid running `npm update` (and `composer update`) like a plague. They tend to do more than what I want. So I heartily agree with including `package-lock` in the git repo of the front end, like we do with the back end.

HEADS-UP: I'm going to push a commit to your branch, with a few CSS changes. Tweak as you see fit:
* Add back top shadow to the download buttons; remove the background color change.
* Add the border and shadow to the "key" button in the login dialog. For these small buttons I feel that they need the shadow both on the top and the left or they blend in with the background. I also made the button a little larger or it looked ridiculous with the full border.
* Add the border and shadow to the "copy to clipboard" buttons in "direct links". Remove the bottom compensation because at least for me it looks unaligned with that compensation in place?
* Add the border and shadow to the token dialog buttons (here the compensation looks just fine, BTW).

This should take care of the dialog boxes. There are more "unfinished" buttons in Settings, Users, etc., but that's a separate problem.

kamil4 (approved):
I finished going over my notes from the code review of this PR and I'm satisfied with its state at this point (though see two notes about missing localizations that I just added).

scripts/main/upload.js
```diff
@@ -132,19 +242,20 @@ upload.start = {
...
 				upload.notify(lychee.locale["UPLOAD_COMPLETE"]);
 			} else {
 				// Error
-				showCloseButton();
+				upload.showProgressReportCloseButton();
 				if (shallCancelUpload) {
-					$(".basicModal .rows .row:nth-child(n+" + (latestFileIdx + 2).toString() + ") .status")
-						.html(lychee.locale["UPLOAD_CANCELLED"])
-						.addClass("warning");
+					const row = upload.buildReportRow("General");
```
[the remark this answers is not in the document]
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/069b9065850e8f37d0d61c4b5a7eca067ee7e0cc.

scripts/main/upload.js
```diff
@@ -571,44 +718,71 @@ upload.start = {
...
 							} else {
-								// If we do not have a current row yet, we
-								// simply append it to the list of rows
-								// (this might happen if the event occurs
-								// before the first progress report)
-								jqEventRow = $(build.uploadNewFile(report.path || "General")).appendTo(jqRows);
-								topSkip += jqEventRow.outerHeight();
+								// The event report does not refer to a
+								// specific directory.
+								row = upload.buildReportRow("General");
```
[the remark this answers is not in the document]
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/322/commits/069b9065850e8f37d0d61c4b5a7eca067ee7e0cc.

PR #323, opened 2022-09-18
Merge `gallery`, `view` and `frame` Into a Unified Frontend

nagmat84:
This PR merges the three distinct frontend modes "gallery", "view" and "frame" into a single unified framework. In doing so, many code duplications and "sub implementations" of JS objects are thrown out.

This PR has taken the different Blade views of the backend and merged the resulting HTML markup into a new HTML file [`frontend.html`](https://github.com/LycheeOrg/Lychee-front/blob/a6eea57cc441bf5bf99b23ec9f6bd83135986bca/html/frontend.html).

The view mode is now accessible via `#view/<photo-id>`, the frame mode via `#frame' . Redirections from the old URLs are enabled via the backend.

**A note and disclaimer**

I am not very fond of this PR myself either. Basically, it is an intermediate step with many things left to do. In particular, the new "root" HTML file `frontend.html` puts the three different modes as independent `<div>` elements one after another and CSS ensures that only one of these elements is visible at a time. This is not my ultimate goal.

Independent of the [current security issue](https://github.com/LycheeOrg/Lychee-front/security/advisories/GHSA-cr79-38hg-27gv) I developed some ideas how to re-structure the whole frontend and improve on its (currently non-existing) modular design. In order to not accidentally break things for the view and frame mode, I wanted to take those into account right from the beginning. This implied that they needed to be merged first.

In order to avoid a single "big bang" PR which takes years to review, I wanted to get this merge out of the way with the least amount of lines of newly added code as possible and with the least amount of change as possible. Unfortunately, this also means that there had to be some nasty workarounds.

**Addendum**

The failing CI is caused by a moved file which has otherwise not been changed by this PR. The CI complains about a poor font definition which the CI considers to be "new".


package.json
```diff
@@ -15,8 +15,8 @@
 	"scripts": {
 		"start": "gulp watch",
 		"compile": "gulp",
-		"check-formatting": "prettier --check .",
-		"format": "prettier --write ."
+		"check-formatting": "prettier --check scripts/ styles/",
+		"format": "prettier --write scripts/ styles/"
```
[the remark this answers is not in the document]
nagmat84: Because at some point the prettier was whining about code in `/node_modules`, `/deps` and `/modules`. The prettier should only check our own code, not others people code which we use as a dependency. Moreover `.` also checks the files on the rool level such as `package-lock.json` which is an auto-generated file by `npm update`.

Moreover, efficiency. The prettier should not even parse the whole `/node_modules` with thousands of files even if it does not find anything.

scripts/main/lychee.js
```diff
@@ -383,23 +561,42 @@ lychee.parsePublicInitializationData = function (data) {
...
+	lychee.site_owner = data.config.site_owner;
+	lychee.site_copyright_begin = data.config.site_copyright_begin;
+	lychee.site_copyright_end = data.config.site_copyright_end;
+
+	lychee.footer_show_social_media = data.config.footer_show_social_media === "1";
+	lychee.footer_show_copyright = data.config.footer_show_copyright === "1";
+	lychee.footer_additional_text = data.config.footer_additional_text;
+
+	lychee.mod_frame_enabled = data.config.mod_frame_enabled === "1";
+	lychee.mod_frame_refresh = Number.parseInt(data.config.mod_frame_refresh, 10) || 30;
```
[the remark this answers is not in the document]
nagmat84: Totally agreed from a software architecture perspective. Actually, this whole code should not even need to exist at all.

But there was a reason why I kept it like that. I wanted to keep this PR "small" in the sense that I only wanted to merge the previously distinct modes "view", "frame" and "gallery" into one and otherwise change as little as possible and only "move" code around but not actually change anything.

I hoped that this approach might speed up the review process. (I would like to here @kami4 opinion about that.) As I said, this PR is not supposed to be the last one or even be "good" by any standards. The frontend is complete spaghetti code.

My final intention is to completely clean up the architecture of the frontend eventually (hopefully the week after the next, if time permits). Then, I would like to remove strange JS objects like `build` or `visible` which do not make any sense to me, but only have objects like `ViewManager`, `AbstractView`, `MapView`, `AlbumView`, `PhotoView` and `AlbumController`, `PhotoController`, etc. For example `visible.photo` would then be moved to `PhotoView.isVisible`.

However, I feared if I would start with the step above as the first step for the "gallery" mode only, I might have accidentally break the "view" and the "frame" mode, because they are the unbeloved children. Hence, I decided to take the following approach:

 1. Preparation 1:
      - Merge "gallery", "frame" and "view" (this PR)
      - Move code around as necessary and throw out duplicates, but don't change anything on the fundamental level
 3. Preparation 2: Revamp the box model of the frontend (see https://github.com/LycheeOrg/Lychee-front/pull/335)
 4. Clean up the architecture step-by-step and many individual PRs

Yes, I could have done everything in one "big bang", but then I feared that the PR would become huge and reviews would take forever.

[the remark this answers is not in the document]
nagmat84: I would like to keep this conversation unresolved, until @kamil4 gave his opinion.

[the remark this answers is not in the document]
kamil4: I'm sort-of wondering myself if many small PRs vs one big one improve the overall throughput, but for sure they cut down on latency. I'm generally in favor of smaller PRs.

scripts/main/lychee.js
```diff
@@ -875,38 +1120,68 @@ lychee.setMode = function (mode) {
...
+		$(".header__search, .header__clear", ".header__toolbar--public").hide();
 
 		if (!lychee.editor_enabled) {
-			$("#button_rotate_cwise").remove();
-			$("#button_rotate_ccwise").remove();
+			$("#button_rotate_cwise").hide();
+			$("#button_rotate_ccwise").hide();
 		}
 		return;
 	} else {
```
kamil4, line 1147:
```suggestion
	}
```

scripts/main/lychee.js
```diff
@@ -875,38 +1120,68 @@ lychee.setMode = function (mode) {
...
-		$(".header__search, .header__clear", ".header__toolbar--albums").remove();
-		$("#button_rotate_cwise").remove();
-		$("#button_rotate_ccwise").remove();
+		$(".header__search, .header__clear", ".header__toolbar--albums").hide();
+		$("#button_rotate_cwise").hide();
+		$("#button_rotate_ccwise").hide();
 	}
```
kamil4, line 1151:
```suggestion
	$(".header__search, .header__clear", ".header__toolbar--albums").hide();
	$("#button_rotate_cwise").hide();
	$("#button_rotate_ccwise").hide();
```

Considering that the body following the `if` condition above unconditionally returns, the `else` here is very confusing, especially since it doesn't extend through the rest of the method (yes, I know that the code had already been structured that way before).

scripts/main/lychee.js
```diff
@@ -875,38 +1120,68 @@ lychee.setMode = function (mode) {
...
+			bodyClasses.replace("mode-view", "mode-frame");
+			bodyClasses.replace("mode-gallery", "mode-frame");
+		}
+		Mousetrap.unbind(["esc", "command+up"]);
+
+		$("#button_back, a#next, a#previous").hide();
+		$(".no_content").hide();
 
 		lychee.publicMode = true;
 		lychee.viewMode = true;
```
kamil4, line 1187: Should `viewMode` really be set to `true` here?

styles/main/_imageview.scss
```diff
@@ -5,16 +5,12 @@
...
+	background-color: #0f0f0f;
 	transition: background-color 0.3s;
 
 	// Modes -------------------------------------------------------------- //
-	&.view {
-		background-color: inherit;
-	}
 
 	&.full {
-		background-color: black(1);
```
kamil4, line 17: I will admit that I always found it rather peculiar that the background in the photo view was ever-so-slightly transparent (and only initially; it would turn into pure, non-transparent black after a few seconds once the header disappears).

So I don't really mind that you removed that transparency. I'm wondering though why you decided to use a lighter shade for the background (`0f` is not the same as `10`, is it?) and why we aren't going full-black in `full` mode anymore.

I haven't tried it, but personally I would be inclined to simply use full-black always in photo view and be done with it.

styles/main/_mapview.scss
```diff
@@ -5,16 +5,12 @@
...
+	background-color: #0f0f0f;
 	transition: background-color 0.3s;
 
 	// Modes -------------------------------------------------------------- //
-	&.view {
-		background-color: inherit;
-	}
 
 	&.full {
-		background-color: black(1);
```
kamil4, line 17: See my comment for `_imageview.scss`.

styles/main/main.scss
```diff
@@ -60,9 +60,6 @@ body {
 	-webkit-font-smoothing: antialiased;
 	-moz-font-smoothing: antialiased;
 	-moz-osx-font-smoothing: grayscale;
-	&.view {
-		background-color: #0f0f0f;
```
kamil4, line 64: Oh, I see now where `0f` came from... Still, I don't like the rare `view` mode taking precedence over a lot more common regular photo view.

kamil4 (changes requested):
I finished my initial read through this PR. I still need to see the server side and I want to test it as there are a number of things I want to try...

scripts/main/lychee.js
```diff
@@ -744,6 +972,12 @@ lychee.load = function (autoplay = true) {
 			if (visible.sidebar()) sidebar.toggle(false);
 			mapview.open();
 			lychee.footer_hide();
+		} else if (albumID === "frame") {
+			if (lychee.mod_frame_enabled) {
+				frame.initAndStart();
+			} else {
+				loadingBar.show("error", "Frame mode disabled");
```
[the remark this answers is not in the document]
nagmat84: We are inconsistent about that (as always). We localize many error messages but it is far from being complete. In particular, strings which come from the "view" and the "frame" mode (as this one) are mostly not localized, probably because these two modes have always been the unbeloved children.

Here, we have a comparable issue like in other places, too. I only moved this code from the former frame mode into the unified frontend, hence this code appears in the diff. I could have fixed the localization, but I did not feel like doing so. Otherwise I also should have fixed dozens of other strings, too, for the sake of consistency. Maybe, this should be an independent PR where we step over all strings (it shouldn't be too difficult to `grep` for then) and then add localizations for all of them.

What do you want? Do it now? Do it later?

scripts/main/lychee.js
nagmat84: I am not sure. At least there does not seem to be a technical necessity for it. The variable `lychee.viewMode` is only queried in three places and none of them belongs to code which is executed during frame mode. Hence, the value does not matter.

However, I did it that way to keep the code consistent with the code in the `else if`-branch for the view mode earlier (13 lines above). See
```php
} else if (mode === "view") {
	if (!bodyClasses.contains("mode-view")) {
		bodyClasses.replace("mode-none", "mode-view");
		bodyClasses.replace("mode-frame", "mode-view");
		bodyClasses.replace("mode-gallery", "mode-view");
	}
	Mousetrap.unbind(["esc", "command+up"]);

	$("#button_back, a#next, a#previous").hide();
	$(".no_content").hide();

	lychee.publicMode = true;
	lychee.viewMode = true;
} else if (mode === "frame") {
	if (!bodyClasses.contains("mode-frame")) {
		bodyClasses.replace("mode-none", "mode-frame");
		bodyClasses.replace("mode-view", "mode-frame");
		bodyClasses.replace("mode-gallery", "mode-frame");
	}
	Mousetrap.unbind(["esc", "command+up"]);

	$("#button_back, a#next, a#previous").hide();
	$(".no_content").hide();

	lychee.publicMode = true;
	lychee.viewMode = true;
}
```

Except for the CSS classes on the `<body>` both branches are identical. In fact, it is also unnecessary to hide the buttons `a#previos` and `a#next` in frame mode, because these buttons will be completely covered by the background of the frame anyway. Hence, the last `else-if`-branch may even be simplified further. The only essential aspect is to unbind the "ESC" key.

However, in my perspective the view mode and the frame mode have a lot in common. They both show a large photo in the center and disallow most interaction with a user. In the long run both modes might share more code (for the image in the center) than they do at the moment. In order to stress that similarity, I considered it helpful to keep both branches as closely together as possible. That is why the code is as it is. But of course this is arguable.

Do you want me to remove that code?

styles/main/main.scss
nagmat84: Right, this is the source of that background color. In order to keep the discussion at one place, I will close this conversation here. We should continue the discussion in  https://github.com/LycheeOrg/Lychee-front/pull/323#discussion_r1001350913.

styles/main/_mapview.scss
nagmat84: In order to keep the discussion at one place, I will close this conversation here. We should continue the discussion in  https://github.com/LycheeOrg/Lychee-front/pull/323#discussion_r1001350913.

styles/main/_imageview.scss
nagmat84: > So I don't really mind that you removed that transparency.

Yes, I removed the transparency on purpose. A transparency of only 2% is not good. If one really wanted the background to shine through, one would need transparency values between 40% and 60%. We do so for the background of the photo view on TVs. There the album and all its photos are still visible in the background of the photo.

I am under the impression that these small values of transparency are not really meant as a transparency effect, but are an exploit to darken or lighten some base color. The CSS currently uses this approach for certain font colors, too, where the font color is `#ffffff` with 2-5% transparency instead of something like `#f0f0f0`. But this only works with the implicit assumption that the original background is dark.

> `0f` is not the same as `10`

No, it isn't. However, as 10h = 0fh + 1 holds, I doubt that anybody can actually see the difference on a monitor. But of course, it is not good at all to use two different color codes for semantically identical things.

At the moment, we are essentially using three different colors for backgrounds:

 1. the background color for albums listings
 2. the background color for photo views (in non-full mode)
 3. the background color for photo views (in full mode)

(In fact, there are probably 20 different colors due to minor differences in the values as the one above, but I assume these are just oversights due to the spaghetti-nature of the CSS.) On top, there are two more background colors

 - the background color of the map view
 - the background color of the frame view

However, I would call these two background colors only of theoretical concern as these backgrounds should never be visible in practice. The map completely covers the background and the frame uses a blurred version of the photo as an additional background layer. Hence, these background colors are only visible if something fails, e.g. a tile of the map fails to load, etc.

*Background color for album listings*

Screenshot_20221021_125624

It is defined as `#1d1d1d` on the `<body>` even though the color picker reports it as `#1c1c1c` between the photos. (Probably due to some other blending effect.)

*Background color for photo views (in non-full mode)*

That is a complicated case:

Screenshot_20221021_130605

Screenshot_20221021_130633

The background color of the `<div id="imgageview">` is (currently) defined as `rgba(10, 10, 10, 0.98)` with that unfortunate 2% transparency. However, the color is not even applied as the box for the image covers most of the background due to its symmetric margins on the left and the right (see 2nd screenshot) and the actual visible color turns out to be pure black `#000000`.

*Background color for photo views (in full mode)*

Easy case again, it is pure black `#000000`.

Screenshot_20221021_130715

> and why we aren't going full-black in full mode anymore.

That has to be an oversight on my sight.

> I haven't tried it, but personally I would be inclined to simply use full-black always in photo view and be done with it.

OK. Yes, I agree that there should be at most two colors for any kind of background: the dark anthracite (`#1d1d1d`) and black (`#000000`). Even though it is actually beyond the scope of this PR, I will define two SCSS colors for that and then apply them everywhere.

I assume you mean that the background color for photos shall always be black, i.e. in non-full and full mode, right? As this is the current behavior, I am inclined to do it this way as this option does not require any team discussion.

But personally, I have always been disturbed by the change of the background color between album view and (normal, non-full) photo view. Personally, I would prefer the anthracite background color in photo view, too, and only switch to black in full mode.

nagmat84: Fixed (hopefully) everything in: https://github.com/LycheeOrg/Lychee-front/pull/323/commits/2d00083cb2ed51c3e5ea56542f5dd97ad15f6903

scripts/main/frame.js
```diff
@@ -0,0 +1,221 @@
...
+	 */
+	const onError = function (jqXHR, params, lycheeException) {
+		const msg = jqXHR.statusText + (lycheeException ? " - " + lycheeException.message : "");
+		console.error("The server returned an error response", {
+			description: msg,
+			params: params,
+			response: lycheeException,
+		});
+		alert(msg);
+		frame.stop();
```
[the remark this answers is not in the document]
nagmat84: >  the alert box seems like a bad idea (I realize it was there before this refactoring).

I know exactly what you mean.

Yes, I kept it like this as it was this way before. I guess the frame mode used the alert box (instead of the disappearing error bar which Lychee provides), because the error bar was not available in frame mode until now. After https://github.com/LycheeOrg/Lychee-front/pull/335 has been merged, we can consider to use that error bar. Currently, it won't do anything good (even if it available now), because the background box of the frame mode still covers the error bar.

We could remove the alert box and add a "TODO" comment instead which hopefully remembers us to use the error bar once it will become possible or we leave it this way for now. I don't mind either way.

> I'm wondering if stopping the loop is the right thing to do?

For me this is the logical consequence of showing an alert box. The alert box will hide most of the image anyway and need to be clicked away explicitly. Hence, continuing the loop is not very helpful.

For me this is a both or nothing decision. If we keep the alert box, we should also stop the loop. If we remove the alert box, we can try going on.

@kamil4 It's your decision.

scripts/main/lychee.js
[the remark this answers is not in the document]
kamil4: Well, neither the view nor the frame mode supported localization as far as I know, so no surprise.

I thought @qwerty287 did such a grep relatively recently and added the missing strings he could find?

I'm generally for trying not to add more issues with the code we add. But, as you said, this is only the first PR of a series. So if you address this in one of the subsequent PRs, I'm OK with that.

scripts/main/frame.js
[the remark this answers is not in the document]
kamil4: I recommend to comment out the alert box, add a TODO, and let the loop continue.

scripts/main/lychee.js
kamil4: I have indeed wondered why we even try to hide some buttons that won't be visible anyway. We do need to make sure that the keyboard shortcuts don't work, but other than that...

But since the code is already there, I'm not overly bothered by it. It just seems... redundant.

scripts/main/lychee.js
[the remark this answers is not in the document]
nagmat84: Personally, I don't feel like addressing it. (I hate these "cleaning up" jobs.) But yes, I probably will. I just did a grep for `loadingBar.show\("\w+", "|\.textContent = "` and found four untranslated strings. Probably, there are some more if one searches for other keywords as well.

Yes, I will do that in another PR.

scripts/main/frame.js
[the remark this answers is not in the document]
nagmat84: Fixed it in https://github.com/LycheeOrg/Lychee-front/pull/323/commits/3b3de969ef0764bbad4376a4d8ef38db2805a3dd.

scripts/main/lychee.js
kamil4: OK, so now that I tried this in person, I think we definitely need to do more with the view mode, because when I'm logged in, I'm seeing a whole lot of extra buttons:

Some of them work (rotation), some don't (starring). I think they should all be removed. Even when I'm not logged in, the little triangle to the right of the title is still there, and it shouldn't.

styles/main/_imageview.scss
kamil4: > No, it isn't. However, as 10h = 0fh + 1 holds, I doubt that anybody can actually see the difference on a monitor.

Well, I think it was `10` in decimal, so 5 steps of difference, which definitely would be more visible.

> But personally, I have always been disturbed by the change of the background color between album view and (normal, non-full) photo view. Personally, I would prefer the anthracite background color in photo view, too, and only switch to black in full mode.

I'm fine with that, I think. Do you want to do that?

scripts/main/lychee.js
```diff
@@ -875,38 +1120,67 @@ lychee.setMode = function (mode) {
...
+			bodyClasses.replace("mode-frame", "mode-gallery");
+			bodyClasses.replace("mode-view", "mode-gallery");
+		}
 		// After login the keyboard short-cuts to login by password (l) and
 		// by key (k) are not required anymore, so we unbind them.
 		Mousetrap.unbind(["l"]).unbind(["k"]);
 
 		// The code searches by class, so remove the other instance.
-		$(".header__search, .header__clear", ".header__toolbar--public").remove();
+		$(".header__search, .header__clear", ".header__toolbar--public").hide();
```
kamil4, line 1140: So I agree that replacing `remove` with `hide` is A Good Thing in most of these cases. The one where I'm not so sure of is the `.header__search`, because we actually have _two_ of these: one in the public mode header and one in the albums view header. That's why we used to remove the other -- to make sure that the selectors pick only one. Oddly, quick testing I just did on the branch shows no obvious issues though -- I'm not sure though if that's by design or by luck. Your thoughts?

kamil4 (commented):
One other difference I noticed when viewing the network traffic was that (unsurprisingly I suppose) the view and frame mode now call `Session::init` where I don't think they used to do it before. There are good side effects of that (localization) but do we expect this to cause any problems?

scripts/main/lychee.js
nagmat84: That should definitely do not happen. And it did not happen when I finished my work on this PR. I fear this is another victim of the many merges and rebases to keep up with master. Sometimes I really hate Git as it is too brave sometimes. It merges things "successfully" but in fact some outdated, wrong, etc. pp. code slips in.

I need to check what went wrong here.

styles/main/_imageview.scss
nagmat84: > I'm fine with that, I think. Do you want to do that [=change background to anthracite]?

At some point maybe, but not now. This would be a change to the current behavior which should be discussed in the team and not slip in as part of a seemingly unrelated PR.

html/frontend.html
```diff
@@ -0,0 +1,138 @@
+<!DOCTYPE html>
+<html lang="">
```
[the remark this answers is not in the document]
nagmat84: I haven't observe this problem yet, but I think it does not harm to add it. I will do it.

scripts/main/lychee.js
nagmat84: Again, I am totally puzzled. I know that I had some problems with that search input. However, it is too long ago to remember exactly. There are two possibilities and I cannot say which one is true.

Either I had run into the kind of problem which you raise here during my own tests, I had reverted that change for this very reason back to removing elements again and now GIT merge fucked up again.

Or `remove` was actually the problem, because it did not allow you to go back from view mode into normal gallery mode. With the new frontend there is no full page reload between view mode and gallery mode, hence all removed elements are gone forever. It might be that this was the reason why I decided against removing any elements.

But all I have in my memory is a black hole. I know that the search input took me hours to make it right in all modes even after transition from one mode into another.

>  Oddly, quick testing I just did on the branch shows no obvious issues though -- I'm not sure though if that's by design or by luck. Your thoughts?

I would say "by design", because I only now that I was spending hours on it. But it is a fragile design for sure. However, I am not entirely sure, if the current state we are looking at is indeed the correct state. I stopped trusting GIT. But if you say it works, I would say the former is correct although I cannot tell you anymore _why_ it is correct.

-------------------------

Heads-up for Subsequent PRs

Even if it works the design above is extremely fragile. Hence, these are my ideas for another PR. (No need to read it, that is just a heads-up.)

Whoever implemented the frontend seems not to have been very precise in using the `id` attribute and the `class` attribute of HTML tags. At many places the `class` attribute seems to be exploited as an ID (that way is actually works, although inefficiently) and sometimes the `id` attribute is misused as a class. The former actually works, although inefficiently, while the latter definitely violates the HTML specification, because an `id` must be unique. May IDE is yelling at me for that violation. The toolbar is very messy in this particular aspect.

My (medium-term) plans are to completely revamp the toolbar. If you have a look at the follow-up PR https://github.com/LycheeOrg/Lychee-front/pull/335 I already did so for the larger boxes. This means I gave proper IDs to each box such as `lychee_left_menu_container`, `lychee_sidebar_container` to uniquely identify them but pulled out common layout traits into CSS classes such as `.vflex-container`, `.hflex-container`, `.vflex-item-rigid`, etc.

I would like to do something similar for the toolbar, in particular completely reorganize its hierarchical structure. It is not only the search input field which we have twice, the same holds for the back button on the left side of the toolbar, and the title section in the middle of the toolbar. Every toolbar (for albums, album, photo, map) provides a copy of the back button and a title in the middle. We don't see that here in this particular piece of code, because we don't need to hide them. But this is technically wrong and my IDE yells at me for that.

Hence, my mid-term idea is as follows:

 1. Have a single, global `lychee_toolbar_container` with a single, global `lychee_toolbar` inside
 2. Divide the single,  global `lychee_toolbar` into three sections
      1. A single, global left section `lychee_toolbar_lsection` (working title!) which flushes its elements to the left
      2. A single, global middle section `lychee_toolbar_msection` (working title!) which centers its elements
      3. A single, global right section `lychee_toolbar_rsection` (working title!) which flushes its elements to the right
 3. The `lychee_toolbar_lsection` contains
       - a single, global back button,
       - a single, global login button,
       - a single, global settings button
 4. The `lychee_toolbar_msection` contains
       - a single, global album tree button (i.e. this button with the downward pointing arrow)
       - a single, global title element
 5. The `lychee_toolbar_rsection` contains
      - a single, _global_ search input field
      - another container `lychee_action_buttons_container` (working title!)

All the elements above should only exist _once_ and are under the control of something which I currently call `ViewManager` in my head. The `ViewManager` is responsible
 - for loading the individual views (e.g. `PhotoView`, `AlbumView`, `MapView`) and
 - for managing the elements above
Each view would only be in charge of the content of `lychee_action_buttons_container` which it can fill with its individual action buttons.

This approach would ultimately solve the problem of all these duplications we have right now.

html/frontend.html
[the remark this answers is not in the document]
nagmat84: I went through the code one again and I assume we are good even without it.

Previously, we had
```html
<title>{{ $title }}</title>
@yield('head-js')
@yield('head-css')
<link rel="shortcut icon" href="favicon.ico">
```
(see [`resources/views/includes/head.blade.php`](https://github.com/LycheeOrg/Lychee/blob/b4cbfc4fe4aeab21b95ad14df666960ac2b2be9a/resources/views/includes/head.blade.php#L2-L7)). Hence, we started to load (and execute) or JS code before the CSS styles had been parsed. Now, the JS sources are included after the `<body>`.

But more importantly we now have the CSS classes `mode-none`, `mode-gallery`, `mode-view` and `mode-frame` for the `<body>` tag and the initial class for `<body>` is `mode-none`. This hides all potentially problematic parts such as the header. For example, we have the following CSS rule for header

```scss
body.mode-frame,
body.mode-none {
	.header {
		display: none;
	}
}
```

Unless the JS code in `lychee.load` has been executed and activated a particular mode, nothing is visible. Hence, ` $("html").css("visibility", "visible");` would be redundant. I would like to not add it, not only due to its redundancy, but also because I would like to avoid any direct CSS rule on an element.

scripts/main/lychee.js
nagmat84: I fixed the problem (hopefully) in https://github.com/LycheeOrg/Lychee-front/pull/323/commits/acd3b710e306f95aded8c643e2df7d6ed6669966.

I also remember the initial problem which has led to the previous unfortunate situation.

Initially, when the "view" mode had been implemented separately, the view mode provided a toolbar called `header__toolbar--public`. This toolbar contained the share and info button for a single photo. However, this name had already been taken in "gallery" mode for the toolbar for public albums. Moreover, the interdependent view mode defined the toolbar as
```html
<div class="header__toolbar header__toolbar--public header__toolbar--visible">
	<a class="header__title"></a>
	<a class="button button--share" id="button_share" title="Share Photo"><svg class="iconic ionicons"><use link:href="#share-ion"></use></svg></a>
	<a class="button button--info" id="button_info" title="About Photo"><svg class="iconic"><use xlink:href="#info"></use></svg></a>
</div>
```
but at the same time, the gallery mode contained a toolbar `header__toolbar--photo`
```html
<div class="header__toolbar header__toolbar--photo">
	<a class="button" id="button_back"><svg class="iconic"><use xlink:href="#chevron-left" /></svg></a>
	<a class="header__title"></a>
	<!-- ... -->
	<a class="button button--share" id="button_share"><svg class="iconic ionicons"><use xlink:href="#share-ion" /></svg></a>
	<a class="button button--info" id="button_info"><svg class="iconic"><use xlink:href="#info" /></svg></a>
	<!-- ... -->
</div>
```
which (among other) defined two buttons with the same IDs `button_share` and `button_info`. This was the point when I decided to drop the previously independent toolbar `header__toolbar--public` and only use `header__toolbar--photo`.

> Some of them work (rotation), some don't (starring). I think they should all be removed.

In the long run, I would object. IMHO, there should not be an independent view mode at all, but simply a single photo view whose toolbar adopts to the actions which are currently available and takes the authenticated user into account. I thought it would be nice to already push things into that direction, but it isn't that easy, because certain actions don't work in view mode and it is not possible to make them work without refactoring a lot more. This is something which can only be implemented after https://github.com/LycheeOrg/Lychee-front/issues/343.

For the time being, I agree that the non-functional buttons should simply be removed. This is the easiest solution for now. I "hacked" the necessary code into `header.setMode`. This way, "view mode" and "photo mode" still use the same toolbar, i.e. `header__toolbar--photo`, and we don't have the problem of duplicated IDs, but we hide buttons if `lychee.publicMode` is set. This is not a great solution, but requires the least lines of additional code. Moreover, we already do similar things for the album and map toolbar, hence it is even somewhat consistent.

scripts/main/settings.js
```diff
@@ -606,9 +606,12 @@ settings.openTokenDialog = function () {
 
 		updateTokenDialog();
 
-		copyTokenButton.addEventListener(lychee.getEventName(), onCopyToken);
-		resetTokenButton.addEventListener(lychee.getEventName(), onResetToken);
-		disableTokenButton.addEventListener(lychee.getEventName(), onDisableToken);
+		copyTokenButton.addEventListener("click", onCopyToken);
+		copyTokenButton.addEventListener("touchend", onCopyToken);
```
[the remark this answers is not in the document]
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/323/commits/46704a9797fbe3382af7981d1400fcd72b02705d.

html/frontend.html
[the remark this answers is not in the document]
kamil4: OK! We should remove that one line from `init.js` then, right?

scripts/main/lychee.js
kamil4: Agreed. We should make the photo mode work in both cases.

html/frontend.html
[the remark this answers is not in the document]
nagmat84: Sure. Fixed in https://github.com/LycheeOrg/Lychee-front/pull/323/commits/4a0180341721d03b6f6a0e82e601a04667080ca2.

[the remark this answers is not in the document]
nagmat84: Unfortunately, I was wrong. We need it.

My error was to assume that the icons of the header are also added dynamically to the toolbar. Then my argument about the JS being loaded last would have hold. But the icons are statically coded into the main HTML markup. This may lead to a situation where the browser starts rendering the HTML page (incl. the icons), but the CSS is not yet loaded. Also the icons are part of the main file, but the CSS is included via `<link>` and hence the CSS is asynchronously loaded after the main file.

> The reason was that at least sometimes when loading Lychee, the page would start to render before CSS was loaded

One can (nearly deterministically) provoke that effect, if one artificially throttles the bandwidth and increases the delay via the developer tools of the browser. In Firefox, the preset "2G regular" did the trick.

I add a fix similar to the old one again in https://github.com/LycheeOrg/Lychee-front/pull/323/commits/3a17fe5829f2a32fae74aefa1abb84bb633afd01.

I wanted to avoid inline styles, but for the moment the old workaround is the best.

kamil4:
Something's definitely broken with the header buttons in the photo view in public mode. I don't see the full screen button, the more menu (with the download button)... On the other hand, the share button seems to be always there even if it shouldn't...

nagmat84:
Sooooorry. :cry: (And also sorry for not responding in timely manner.) I made an attempt to fix it a second time in https://github.com/LycheeOrg/Lychee-front/pull/323/commits/6e395bc6400a0f2c951410e820d118397a2235c3. (I also synced the frontend into the backend https://github.com/LycheeOrg/Lychee/pull/1522, hence you should probably checkout that.)

I tested both ways to display a photo, i.e. in gallery mode via `#<album-id>/<photo-id>` and in view mode via `/view?p=<photo-id>` for anonymous and admin user. I assume we should be fine now.

Note, in contrast to the current master branch you should now see some more buttons in "view mode" if a user is logged in. On current master, the "view mode" only shows the share button and the info button. On this branch, the "view mode" shows the same buttons as the gallery mode does for an unauthenticated user, i.e. the share button, the info button, the map button (if enabled), the fullscreen button (if fullscreen is available) and the more button (if not empty, i.e. if downloading the original size variant is enabled). However, you should only see button which are "safe", i.e. which don't throw an exception if used.

I had to fix the method `album.isUploadable` which was the culprit all the way along. I hope this did not break anything else, but I haven't found anything so far.

scripts/main/header.js
```diff
@@ -479,113 +479,117 @@ header.setMode = function (mode) {
...
+				const e = $("#button_rotate_cwise, #button_rotate_ccwise");
 				e.hide();
 				tabindex.makeUnfocusable(e);
 			}
 
-			if (photo.json && photo.json.hasOwnProperty("is_share_button_visible") && !photo.json.is_share_button_visible) {
+			if (lychee.enable_button_share && photo.json && photo.json.is_share_button_visible) {
 				const e = $("#button_share");
-				e.hide();
+				e.show();
```
kamil4, line 529: As you inverted the condition here (which I'm in favor of, for the sake of consistency) you need to switch the order of `makeUnfocusable`/`makeFocusable` as well...

kamil4 (changes requested):
I spotted one minor issue during the review. I also still need to test it.

kamil4:
I played with it a bit this evening and unfortunately this is still not ready for merging due to the following problem:
* Uploading to the root album (albums view) doesn't work for an admin for me (weirdly, it seems to pass an ID of one of the *tag albums* in the `Photo:add` request?!). Unusually, it works for non-admin users.

While testing, I also noticed some strange behavior around the "Share" button although it looks like that may not be specific to this branch:
* Album owners don't get to see the "Share" button in the header until the album is made public and the "Share button is visible" is checked. This is contrary to the behavior of other options for authenticated album owners (such as the ability to download -- it's always there for owners, irrespective of any settings).
* When the album is made public and the "Share button is visible" is checked, going (as an authenticated owner) to one of the photos inside the album and opening the visibility dialog is supposed to display read-only visibility status, in this case including a checked "Share button is visible". But that button is not checked.

scripts/main/header.js
nagmat84: Fixed by https://github.com/LycheeOrg/Lychee-front/pull/323/commits/8deda77a8701b6253177823587dcf4f5cbd9eb1c.

nagmat84:
The other issue should (hopefully) been fixed by https://github.com/LycheeOrg/Lychee-front/pull/349 on current master. I merged current master into this branch once again.

kamil4:
The share button became visible for me, but the dialog box still wouldn't open. I removed the checks from the code, which I believe were unnecessary since the proper access control is done at the button visibility level. Sorry for polluting your PR with this code but since the discussion started here...

I also figured out my issue with being unable to upload to root. It was due to the relative positions of my browser window (and its contents) and the file selection dialog. My mouse pointer happened to be hovering over a tag album at the time and if that's the case, `album.getID` returns the ID of that album, not `null`!!! This is a frightening but undoubtedly old bug so I guess we should let it be for now...

kamil4 (approved):

nagmat84:
> The share button became visible for me, but the dialog box still wouldn't open. I removed the checks from the code, which I believe were unnecessary since the proper access control is done at the button visibility level. Sorry for polluting your PR with this code but since the discussion started here...

No worries. I'm glad that you checked the PR that thoroughly. It is simply interesting how many bugs suddenly become apparent which probably have been lurking around for quite some while.

No worries about "polluting" this PR. I think the code (not necessarily this PR, but the frontend in general) has reached some that state such that it is hardly possible to pollute it any more. 😂 The whole visibility checks are completely crazy.

> I also figured out my issue with being unable to upload to root. It was due to the relative positions of my browser window (and its contents) and the file selection dialog. My mouse pointer happened to be hovering over a tag album at the time and if that's the case, `album.getID` returns the ID of that album, not `null`!!! This is a frightening but undoubtedly old bug so I guess we should let it be for now...

Thats fits in the overall picture. But this is also an indicator that my recently added comment to the code where the "drop" event handler is registered is more than justified. It is just insane to register the "drop" event handler on the global browser window and then attempt to find out which element really received the drop element. This has to fail. However, this can only be fixed in the long run and only after the box model has been revamped.

ildyria (approved):
Tested. LGTM.

PR #325, opened 2022-09-22
Fix most ecaping of user provided input

nagmat84:
This PR add `lychee.html` and the double-dollar operator in front of user-supplied data.

I did a quick search for the regex `[^$]\$\{[^}]*(title|Title|name|desc)` and hopefully I found all problematic cases. This is a hotfix. In particular it might introduce some undesired regression due to "double" escaping. But it should mostly be fine.


ildyria (approved):

nagmat84:
I converted this to a draft, because after playing around with this PR some while, I encountered the "double-escape" problem. :-(

screenshot-2022-09-24 _17-38-15

PR #326, opened 2022-09-28
Limit HTML escaping to album description

kamil4:
This fixes a regression introduced in #325, where the whole subtitle is escaped, not just the case when it's a user-supplied `description`. In particular, if the `album_subtitle_style` config is set to `takedate`, the subtitle is a `<span>`, which on current master results in eyesores like this:

2-crop

In my opinion the issue being fixed is a release-blocker so this PR should be merged before 4.6.1 is released.

### DIFF
--- a/scripts/main/build.js
+++ b/scripts/main/build.js
@@ -86,7 +86,7 @@ build.album = function (data, disabled = false) {
 	// default: any other type defaults to old style setting subtitles based of album sorting
 	switch (lychee.album_subtitle_type) {
 		case "description":
-			subtitle = data.description ? data.description : "";
+			subtitle = data.description ? lychee.escapeHTML(data.description) : "";
 			break;
 		case "takedate":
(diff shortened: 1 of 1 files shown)


nagmat84:
Nicely, spotted. I warned you about lurking regressions like this one in #325. I played around with the changes I made in #325 but I totally forgot that the photo overlay is configurable.

I am generally fine with this PR and given the fact that I intend to rewrite the views as well (not only the dialogs) this fix will do.

Otherwise I would have insisted on renaming the variable `subtitle` into `htmlSubtitle` to reflect the circumstance that this variable is already supposed to contain proper HTML and does not need a double-dollar operator. In that case we should also have wrapped `htmlSubtitle = formattedMinTs` (and other things) into `escapeHtml` just for safety.

But as I said: As an interim solution I am fine with it.

nagmat84 (approved):

PR #327, opened 2022-10-03
Enable README.md syntax highlight for readability

PeterDaveHello:
Should make the code blocks be easier to be read 🎉 !


d7415 (approved):
Most of them don't do a lot, but it makes sense.

PR #334, opened 2022-10-19
Improve photo layout performance

nagmat84:
While I have been working on some other PRs I noticed that some bottle necks in the layout routine for albums:

- When we rearrange the photos of the layout, the layout should be hidden such that each relayouted child triggers a re-drawing
- Stop querying the DOM for CSS selectors several times
- When the DOM has been manipulated (i.e. the photos have been added), we must provide the engine a time slice to catch up with the new DOM otherwise querying CSS attributes is slow (--> make `justify` asynchronous, wrap it into `setTimeout`).


scripts/main/sidebar.js
```diff
@@ -137,7 +137,7 @@ sidebar.toggle = function (is_user_initiated) {
 		header.dom(".button--info").toggleClass("active");
 		lychee.content.toggleClass("content--sidebar");
 		lychee.imageview.toggleClass("image--sidebar");
-		if (typeof view !== "undefined") view.album.content.justify(album.json ? album.json.photos : []);
```
kamil4, line 140: In case you were wondering "what the hell was that for", it was to prevent the standalone view mode from crashing, I believe. But with #323 I guess that may no longer be needed?

scripts/main/view.js
```diff
@@ -216,9 +221,14 @@ view.album = {
 
 			if (photosData !== "") {
 				if (lychee.layout === 1) {
-					photosData = '<div class="justified-layout">' + photosData + "</div>";
+					// The CSS class 'relayouting' prevents the DIV from being
```
kamil4, line 224: With my "grammar nazi" hat on, I think strictly speaking the class should be called `relayingout`? Personally I think I would drop the `re` as well, i.e., `layingout`...

scripts/main/view.js
```diff
@@ -440,6 +450,26 @@ view.album = {
...
+		 * keep a pointer on an outdated array of photos when {@link album} or
+		 * {@link search} gets cleared.
+		 *
+		 * @param {?{json: ?{photos: Photo[]}}} photoDataSource
+		 */
+		setPhotoDataSource: function (photoDataSource) {
+			view.album.content.photoDataSource = photoDataSource;
+			setTimeout(() => view.album.content.justify(), 0);
+		},
+
```
kamil4, line 472: Do we need this method at all? Search now sets `album` to its result as well...

scripts/main/view.js
```diff
@@ -482,51 +518,76 @@ view.album = {
...
 				});
-				// if (lychee.rights.is_admin) console.log(layoutGeometry);
-				$(".justified-layout").css("height", layoutGeometry.containerHeight + "px");
-				$(".justified-layout > div").each(function (i) {
+				// Temporarily hide the container such that not each
+				// modification of every photo triggers a UI update.
+				jqJustifiedLayout.addClass("relayouting");
+				// We must give the UI worker a chance to apply the class
+				// above, hence the remaining part of this method must be
+				// done asynchronously.
```
kamil4, line 539: Am I missing something? I don't see the reminder of this method being done asynchronously...

kamil4 (commented):
Looks mostly good to me (untested) but I did have a few questions/comments.

scripts/main/sidebar.js
nagmat84: Thanks good to know.

scripts/main/view.js
nagmat84: I don't mind. You are the English pro. I will change it. But maybe "laying-out" with a hyphen for better readability?

kamil4: Fine with me...

scripts/main/view.js
kamil4: Sorry, here's what I meant:

Some time back, you refactored the search code so that search results were stored in a new location: `search.json`. I was worried that too many places may depend on the currently displayed photos being stored in`album.json`, so I requested that you store the search results there as well, effectively creating a virtual "search smart album", so to speak. You implemented that.

Therefore, I don't think there's any reason anymore to decide whether to use `album` or `search` as the source of photos: in any situation where `search` would be appropriate, the same result is also stored in `album`. Therefore, always loading from `album` should be fine. We shouldn't need a method that allows to specify a different source.

Does that make sense?

nagmat84: Yes that makes total sense. I cannot even remember that I implemented that change you are speaking about. So, what you are saying is that the previous lines of code like
```js
if (visible.album()) view.album.content.justify(album.json ? album.json.photos : []);
if (visible.search()) view.album.content.justify(search.json.photos);
```
have already been unnecessary and `view.album.content.justify()` would have sufficed.

That's would be great. I am going to check that and probably change it. (How embarrassing for me to not remembering that.)

scripts/main/view.js
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/334/commits/dcfc2c2dfeeebf03d64fd7c7526002327f3240cd.

scripts/main/view.js
nagmat84: Fixed in https://github.com/LycheeOrg/Lychee-front/pull/334/commits/c7a780b47f15e5aa126db8d633748c8873ea659b.

scripts/main/view.js
nagmat84: @kamil4 You were absolutely right.

Although I coded this

https://github.com/LycheeOrg/Lychee-front/blame/a89efe2b59da434dcf3a933d5a3318a4c1edae5f/scripts/main/search.js#L58-L69

myself, I could not remember.

Fixed it in https://github.com/LycheeOrg/Lychee-front/pull/334/commits/fcb3cfb3775e31c01b09e426296f49c6ac853ea0.

nagmat84:
I had to revert one change in https://github.com/LycheeOrg/Lychee-front/pull/334/commits/c6bcfac33ae9cf0e159874c8a1a89d281a92849d. I also added a (long) comment what the problem is so that I (or someone else) remembers and understands what the root cause is.

scripts/main/view.js
kamil4: No worries; I remembered it for you :smiley:.

kamil4 (approved):
I played with it around a bit and didn't spot any problems.

PR #335, opened 2022-10-19
Refactor box model

nagmat84:
### Disclaimer

This is a follow-up of https://github.com/LycheeOrg/Lychee-front/pull/323 and https://github.com/LycheeOrg/Lychee-front/pull/334. This PR is strictly ahead of those both PRs and hence shows all commits and diffs which are already contained in https://github.com/LycheeOrg/Lychee-front/pull/323 and https://github.com/LycheeOrg/Lychee-front/pull/334. Hence, the diff of this PR will "shrink" after those both PRs have been merged.

### Summary

This PR revamps the box model of the Lychee frontend such that the box model follows the visual layout of the frontend more closely. In doing so it also fixes two long-lasting issues:

 - The wrong position of the vertical scroll bars is fixed
 - The height of the footer is now always correctly taken into account even for "huge" footers.

**Screenshots to illustrate the issues**

_New scrollbars_

As you can see, the scrollable content part of the view and the sidebar have independent vertical scrollbars on their respective right edge.

01 Screenshot_20221019_175158

_Correct height of footer_

If the footer is large (i.e. every possible option is enabled), the last photo is fully visible and the footer is below it

02 Screenshot_20221019_175517

_Previous erroneous behavior on current master_

 - The scrollbar of the content view is at the very outer right while the scrollbar of the right sidebar ist left of it. This is very confusing. In particular it was possible to accidentally scroll the "wrong" area.
 - The lower half of the last photo is covered by the footer (the overlay is not visible anymore)
 - The right half of the footer is covered by the sidebar (you might consider this okay'ish, if you think of the sidebar as an overlay)

03 Previous

### Under the hood

 - New box model, the boxes are now organized as follows
   04 Box Model
   The illustration is only meant as a helper to understand what has changed.
 - The new box model allows to get rid of JS code which previously attempted to dynamically calculate certain sizes (such as the size of the footer) and to get rid of a lot of CSS classes (such as the CSS class which adopted the right padding of the content area depending on whether the sidebar was open or closed). Hence, the new code should be a lot less error-prone.
 - Previously, the `<body>` was the element which (erroneously) contained the vertical scrollbar for the scrollable content area. (That was also the reason, why the scrollbar appeared at the very right side.) Now the respective `<div>` is scrollable; in particular the `<body>` is always fully visible. This also allowed to remove the JS package `scrolllock`. Previously, this module was used to keep certain elements in sight even when the body was scrolled.


ildyria:
> ### Under the hood
> * New box model, the boxes are now organized as follows
>   04 Box Model
>   The illustration is only meant as a helper to understand what has changed.
> * The new box model allows to get rid of JS code which previously attempted to dynamically calculate certain sizes (such as the size of the footer) and to get rid of a lot of CSS classes (such as the CSS class which adopted the right padding of the content area depending on whether the sidebar was open or closed). Hence, the new code should be a lot less error-prone.
> * Previously, the `<body>` was the element which (erroneously) contained the vertical scrollbar for the scrollable content area. (That was also the reason, why the scrollbar appeared at the very right side.) Now the respective `<div>` is scrollable; in particular the `<body>` is always fully visible. This also allowed to remove the JS package `scrolllock`. Previously, this module was used to keep certain elements in sight even when the body was scrolled.

May I request this to be added to Lychee Docs ? :)

ildyria (commented):
Quickly tested. LGTM.

ildyria (approved):

PR #336, opened 2022-10-20
Fix login modal error highlighting

qwerty287:


nagmat84 (approved):
I don't know how this bug slipped in again. Seems to another victim of the many merges and rebases since the PR was written.

Let me grep the source files for other occurences of `basicModel.error`. Maybe there are some more which came back from hell.

nagmat84:
> Let me grep the source files for other occurences of `basicModel.error`. Maybe there are some more which came back from hell.

Did it. They seem the only three occurrences.

PR #337, opened 2022-10-20
Use stackblur from NPM instead of local file

qwerty287:
https://github.com/flozz/StackBlur


nagmat84 (approved):
Makes total sense to me.

I haven't tested it myself, hence my usual question: Did you run it and check whether the frame mode still works as expected, i.e. the margin contains a blurred version of the picture at the center?

PR #338, opened 2022-10-20
Merge `deps`/`modules` folders

qwerty287:


nagmat84:
I wouldn't do this, but simply purge the `deps` folder completely. Actually, there is no need to copy over `basicContext` into the `deps` folder, but we can add it to the NPM dependencies they same way as we do for `basicModal`. A package is not required to be a published NPM module, you can also point to a GIT repository and then add the GIT hash or a version tag after the hash tag `#`. This way we could remove the copy of `basicContext` completely from our code repo.

I don't know what we can do about `Leaflet.Photo-gh-pages`, but mostly because I don't know where it originates from.

PR #339, opened 2022-10-21
Add option to append tags

qwerty287:
Adds a simple "Override" checkbox which will append tags if not checked.

### DIFF
--- a/scripts/main/photo.js
+++ b/scripts/main/photo.js
@@ -812,7 +812,8 @@ photo.editTags = function (photoIDs) {
 			<div class="input-group stacked"><input class='text' name='tags' type='text' minlength='1'></div>
 			<div class='input-group compact-inverse'>
 				<label for="override">${lychee.locale["OVERRIDE"]}</label>
-				<input type='checkbox' id='override' name='override' />
+				<input type='checkbox' id='tag_dialog_override_input' name='override' />
+				<p>${lychee.locale["TAGS_OVERRIDE_INFO"]}</p>
 			</div>
 		</form>`;
 
@@ -848,10 +849,10 @@ photo.editTags = function (photoIDs) {
 /**
  * @param {string[]} photoIDs
  * @param {string[]} tags
- * @param {boolean} override
+ * @param {boolean} shall_override
  * @returns {void}
  */
-photo.setTags = function (photoIDs, tags, override) {
+photo.setTags = function (photoIDs, tags, shall_override) {
 	if (visible.photo()) {
 		photo.json.tags = override ? tags : photo.json.tags.concat(tags.filter((t) => !photo.json.tags.includes(t)));
 		view.photo.tags();
@@ -866,7 +867,7 @@ photo.setTags = function (photoIDs, tags, override) {
 		{
 			photoIDs: photoIDs,
 			tags: tags,
-			shall_override: override,
+			shall_override: shall_override,
 		},
 		function () {
 			// If we have any tag albums, force a refresh.
--- a/scripts/main/photo.js
+++ b/scripts/main/photo.js
@@ -793,7 +793,7 @@ photo.editTags = function (photoIDs) {
 	}
 
 	/**
-	 * @param {{tags: string,override: boolean}} data
+	 * @param {{tags: string, override: boolean}} data
 	 * @returns {void}
 	 */
 	const action = function (data) {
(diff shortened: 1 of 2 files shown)


scripts/main/photo.js
```diff
@@ -803,13 +803,17 @@ photo.editTags = function (photoIDs) {
...
 	};
 
 	const setTagDialogBody = `
 		<p></p>
 		<form>
 			<div class="input-group stacked"><input class='text' name='tags' type='text' minlength='1'></div>
+			<div class='input-group compact-inverse'>
+				<label for="override">${lychee.locale["OVERRIDE"]}</label>
+				<input type='checkbox' id='override' name='override' />
+			</div>
```
nagmat84, line 816: Maybe we should add an explanation here, what happens if the checkbox is not checked. In other words an helping text which says that the tags are appended to the existing tags, if the checkbox is unchecked.

scripts/main/photo.js
```diff
@@ -803,13 +803,17 @@ photo.editTags = function (photoIDs) {
...
+		photo.setTags(photoIDs, newTags, data.override);
 	};
 
 	const setTagDialogBody = `
 		<p></p>
 		<form>
 			<div class="input-group stacked"><input class='text' name='tags' type='text' minlength='1'></div>
+			<div class='input-group compact-inverse'>
+				<label for="override">${lychee.locale["OVERRIDE"]}</label>
+				<input type='checkbox' id='override' name='override' />
```
nagmat84, line 815: Having a short `name` for the input element is fine, because it is only used in the context of this method. However, `id` must be unique across the entire HTML page at any moment, and I fear that an ID `override` runs the risk of not being unique.

I had the problem with the ID `header` once, because an item inside our included SVG used the same ID and suddenly a lot of funny and seemingly unexplainable things were happening.

That is why I used rather long and clumsy IDs at other places like `id='tag_dialog_override_input'`, because that somewhat assures that nothing else used the same ID. For example, we include the Dropbox JS module. Can we guarantee that this does not accidentally contain an element whose ID is also `override`, because you might override files in a dropbox? Unfortunately, the HTML name space is flat.

Luckily, these long IDs are not really needed anywhere anymore, hence it doesn't matter how clumsy they are.

scripts/main/photo.js
```diff
@@ -843,11 +848,12 @@ photo.editTags = function (photoIDs) {
 /**
  * @param {string[]} photoIDs
  * @param {string[]} tags
+ * @param {boolean} override
```
nagmat84, line 851: Same thing here. I would prefer something like
```suggestion
 * @param {boolean} shallOverride
```

nagmat84 (changes requested):

scripts/main/photo.js
```diff
@@ -812,7 +812,8 @@ photo.editTags = function (photoIDs) {
 			<div class="input-group stacked"><input class='text' name='tags' type='text' minlength='1'></div>
 			<div class='input-group compact-inverse'>
 				<label for="override">${lychee.locale["OVERRIDE"]}</label>
-				<input type='checkbox' id='override' name='override' />
+				<input type='checkbox' id='tag_dialog_override_input' name='override' />
+				<p>${lychee.locale["TAGS_OVERRIDE_INFO"]}</p>
```
nagmat84, line 816: Don't inject text into HTML code this way. Have a look at the other dialogs how to do it. Use the `.textContent` property of the paragraph to actually set the text.

nagmat84 (changes requested):

scripts/main/photo.js
nagmat84: Ah, and basically this also holds for the the label. I don't know how I could miss that in the first place.

scripts/main/photo.js
```diff
@@ -793,7 +793,7 @@ photo.editTags = function (photoIDs) {
 	}
 
 	/**
-	 * @param {{tags: string}} data
+	 * @param {{tags: string,override: boolean}} data
```
kamil4, line 796:
```suggestion
	 * @param {{tags: string, override: boolean}} data
```

scripts/main/photo.js
```diff
@@ -843,11 +849,12 @@ photo.editTags = function (photoIDs) {
...
  * @param {string[]} photoIDs
  * @param {string[]} tags
+ * @param {boolean} shall_override
  * @returns {void}
  */
-photo.setTags = function (photoIDs, tags) {
+photo.setTags = function (photoIDs, tags, shall_override) {
 	if (visible.photo()) {
-		photo.json.tags = tags;
+		photo.json.tags = override ? tags : photo.json.tags.concat(tags.filter((t) => !photo.json.tags.includes(t)));
```
kamil4, line 858:
```suggestion
		photo.json.tags = shall_override ? tags : photo.json.tags.concat(tags.filter((t) => !photo.json.tags.includes(t)));
```

kamil4 (changes requested):

ildyria (approved):
LGTM. Untested.

nagmat84 (approved):

kamil4 (approved):
(untested)

PR #340, opened 2022-10-21
Replace local files with NPM dependencies

qwerty287:
Migrate to https://github.com/LycheeOrg/Leaflet.Photo and https://github.com/LycheeOrg/basicContext. This has to wait for https://github.com/LycheeOrg/basicContext/pull/3 because it uses the latest commit from the PR.

Tested, working.

Additional changes:
- remove `require` dependency
    - was unused
    - was responsible for 2 critical security alerts
- remove `spin.js` and `leaflet-spin` (unused)


nagmat84 (approved):
I haven't tested it, but I assume you did.

It looks good. Only to questions out of curiosity, why two more packages than I expected have been removed. I thought, we would need `require` to load the Dropbox JS on demand.

nagmat84:
Ah, I should have read your PR message first. :laughing: 

> Additional changes:
>
>  - remove require dependency
>     - was unused
>     - was responsible for 2 critical security alerts
>  - remove spin.js and leaflet-spin (unused)

Are you sure about the `require`? I thought I had seen it in the context of Dropbox.

nagmat84:
> I tested it, but not the dropbox upload. But if the dropbox upload requires require, it would not work because the files for it are not in the gulpfile.

I don't understand what you want to say. I mean, you have removed `require`, so yes, they are not part of the sources anymore, but they have been until now. I don't get your point.

nagmat84:
I checked the frontend code. The Dropbox part should be fine without `require`. But I knew that I had seen `require` somewhere. We use it for the justified layout in

https://github.com/LycheeOrg/Lychee-front/blob/bf7178e4de8766ca9917b41de3a8ddee2645c68e/scripts/main/view.js#L485-L493

Have you also checked, if justified layout still works? If your photos are layed out unjustified, then this code branch won't be executed. Of course, it might as well be possible, that the Gulp plugin `browserify`, or `babel`, resp., take care of that as `require` now is part of the ES2016 standard and hence an external package might be indeed obsolete.

nagmat84:
It is one of the options how photos in an album can be arranged. You can select between "justified" and "unjustified" in the settings. I am not even sure right now, if "justified" is not even the default.

nagmat84:
Great. I approved the PR already.

PR #345, opened 2022-10-25
Compatible with < ES2021

pchampio:
`replaceAll` was lately introduced with ES2021. `6.72%` of the population does not seems to be compatible with it https://caniuse.com/mdn-javascript_builtins_string_replaceall
This (very simple) PR goal is to allow viewing and browsing albums on older devices/browsers not supporting ES2021 (update [scripts/csrf_protection.js](https://github.com/LycheeOrg/Lychee-front/compare/master...pchampio:Lychee-front:patch-1#diff-620863fd7b0f5cf1cc1bf523a3cba85142f9339b28faaf1c0cf74968aac085b5)).

`replaceAll`  is **replaced** with the older replace function with regex.

For good measure, I've also updated [scripts/main/upload.js](https://github.com/LycheeOrg/Lychee-front/compare/master...pchampio:Lychee-front:patch-1#diff-fc7e54aab2fa327b01b451db95a3be861edf80101e3eca5ddd85fea21ced36fa) even through I have not extensively tested the upload on the older device.

```js
"\\ ".replace(/\\ /g, " ") == "\\ ".replaceAll("\\ ", " ") && "\\".replaceAll("\\ ", " ") == "\\".replace(/\\ /g, " ")
=> true
```

### DIFF
--- a/scripts/main/upload.js
+++ b/scripts/main/upload.js
@@ -942,7 +942,7 @@ upload.start = {
 			// After splitting, the escaped spaces must be replaced by
 			// proper spaces as escaping of spaces is a GUI-only thing to
 			// allow input of several paths into a single input field.
-			data.paths = data.paths.match(/(?:\\ |\S)+/g).map((path) => path.replaceAll("\\ ", " "));
+			data.paths = data.paths.match(/(?:\\ |\S)+/g).map((path) => path.replace(/\\ /g, " "));
 			basicModal.close(false, () => importFromServer(data));
 		};
 
--- a/scripts/csrf_protection.js
+++ b/scripts/csrf_protection.js
@@ -21,5 +21,5 @@ csrf.getCSRFCookieValue = function () {
 	// When we send back the value to the server as part of an AJAX request,
 	// Laravel expects an unpadded value.
 	// Hence, we must remove the `%3D`.
-	return cookie ? cookie.split("=")[1].trim().replaceAll("%3D", "") : null;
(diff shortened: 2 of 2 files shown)


kamil4:
Out of curiosity, what is affected by this? According to https://caniuse.com/mdn-javascript_builtins_string_replaceall, all major browsers have supported it for over 2 years at this point. Is it something embedded?

pchampio:
The good'old  iPhone 6 / 6 Plus (2014) | iPhone 5s (2013).
The fix was simple, that why I think it is worth doing it.
Otherwise I understand using newer feature of the standard lib.

kamil4 (approved):
LGTM

scripts/main/upload.js
```diff
@@ -942,7 +942,7 @@ upload.start = {
 			// After splitting, the escaped spaces must be replaced by
 			// proper spaces as escaping of spaces is a GUI-only thing to
 			// allow input of several paths into a single input field.
-			data.paths = data.paths.match(/(?:\\ |\S)+/g).map((path) => path.replaceAll("\\ ", " "));
+			data.paths = data.paths.match(/(?:\\ |\S)+/g).map((path) => path.replace(/\\ /g, " "));
```
nagmat84, line 945: I checked it out and tested it, as I wasn't sure if the number of escape character was right. It is. I tried uploading several directories with spaces. I works.

nagmat84 (approved):

PR #348, opened 2022-11-05
Fixes #347

### LINKED ISSUE #347: [Regression] Photos in the unsorted album cannot be edited
Photos in the unsorted album cannot be edited. No rotation buttons, tags cannot be added, etc. Only the Admin user can do these modifications.

The https://github.com/LycheeOrg/Lychee/commit/1c8eedb22f59b066d76395421f5b37a168db27e8 broke the functionality.

nagmat84:
Fixes https://github.com/LycheeOrg/Lychee-front/issues/347. The regression was introduced by https://github.com/LycheeOrg/Lychee/pull/1443.

Before https://github.com/LycheeOrg/Lychee/pull/1443 the method `album.isUploadable` relied on the undefinedness of the attribute `album.owner_name` as an implicit indicator wether the album at hand was a smart album. While this is still true, i.e. smart album still don't have an owner, this bug fix adds an explicit check for smart albums instead of implicitly relying on the fact that the previous boolean condition "accidentally" evaluated to `true` for undefined values.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -1527,6 +1527,26 @@ album.apply_nsfw_filter = function () {
  * @returns {boolean}
  */
 album.isUploadable = function () {
+	// If no album is loaded, nobody (not even the admin) can upload photos.
+	// We must check this first, before we test for the admin short-cut.
+	//
+	// Particular photo actions (such as starring/unstarring a photo) assume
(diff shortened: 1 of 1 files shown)


qwerty287 (approved):

ildyria (approved):
LGTM.

PR #349, opened 2022-11-06
Fix various GUI issues

nagmat84:
This PR fixes various GUI which @kamil4 discovered during his tests of https://github.com/LycheeOrg/Lychee-front/pull/323 (see https://github.com/LycheeOrg/Lychee-front/pull/323#issuecomment-1304713804), but which already affect the current master. I hope I fixed all of them without breaking anything else. :see_no_evil:

I also hijacked @ildyria PR https://github.com/LycheeOrg/Lychee/pull/1578 for the backend to sync these fixes. Hence, you can simply checkout https://github.com/LycheeOrg/Lychee/pull/1578 to test the changes.

@kamil4 wrote in https://github.com/LycheeOrg/Lychee-front/pull/323#issuecomment-1304713804

> Uploading to the root album (albums view) doesn't work for an admin for me (weirdly, it seems to pass an ID of one of the tag albums in the Photo:add request?!). Unusually, it works for non-admin users.

Should be fixed by https://github.com/LycheeOrg/Lychee-front/commit/aaa6f380289cc7502d5398ecdd4e1825daf6c804. While the fix itself seems to be easy, I added _a lot_ of comments to warn about the fragility of the code.

> Album owners don't get to see the "Share" button in the header until the album is made public and the "Share button is visible" is checked.

Should be fixed by https://github.com/LycheeOrg/Lychee-front/commit/1286b08b4fbb551ac4e3eb26c8cda6da1bf22132. This bug seems to be around for a very long time now. I don't have any non-admin users for my gallery and even if I had I assume I would not have noticed that bug.

> When the album is made public and the "Share button is visible" is checked, going (as an authenticated owner) to one of the photos inside the album and opening the visibility dialog is supposed to display read-only visibility status, in this case including a checked "Share button is visible". But that button is not checked.

Should be fixed by https://github.com/LycheeOrg/Lychee-front/commit/0e023592409f160974a44fff69b552a61fc41990. This one was a silly oversight when migrating to the new modal dialog.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -1535,9 +1535,24 @@ album.apply_nsfw_filter = function () {
  *  - the method returns `true` for regular albums if and only if the album is
  *    owned by the currently authenticated user.
  *
+ * Note, for the time being this method contains a work-around in case
+ * no album is loaded, but the root view is visible.
+ * Currently, this is necessary, because this method is (erroneously) called
+ * for the root view as well.
+ * In order to determine whether the work-around for the root view needs to
+ * be applied, this method checks if the root view is visible based on the
+ * visibility of the corresponding headers.
+ * Hence, the caller must ensure that the appropriate header is set first
+ * in order to obtain a correct result from this method.
+ *
  * @returns {boolean}
  */
 album.isUploadable = function () {
+	// Work-around in case this method is called for the root view
+	if (visible.albums()) {
+		return lychee.rights.is_admin || (lychee.user !== null && !lychee.publicMode && lychee.rights.may_upload);
+	}
+
 	// If no album is loaded, nobody (not even the admin) can upload photos.
 	// We must check this first, before we test for the admin short-cut.
 	//
--- a/scripts/main/albums.js
+++ b/scripts/main/albums.js
@@ -12,6 +12,26 @@ const albums = {
  */
 albums.load = function () {
 	const showRootAlbum = function () {
+		// DO NOT change the order of `header.setMode` and `view.albums.init`.
+		// The latter relies on the header being set correctly.
+		//
+		// `view.albums.init` builds the HTML of the albums view (note the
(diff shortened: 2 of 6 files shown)


scripts/main/album.js
```diff
@@ -1535,9 +1535,24 @@ album.apply_nsfw_filter = function () {
...
+ * visibility of the corresponding headers.
+ * Hence, the caller must ensure that the appropriate header is set first
+ * in order to obtain a correct result from this method.
+ *
  * @returns {boolean}
  */
 album.isUploadable = function () {
+	// Work-around in case this method is called for the root view
+	if (visible.albums()) {
+		return lychee.rights.is_admin || (lychee.user !== null && !lychee.publicMode && lychee.rights.may_upload);
```
ildyria, line 1553:
```suggestion
		return lychee.rights.is_admin || lychee.rights.may_upload;
```

Why is the following not sufficient?

nagmat84: I don't know.

I guess in the long run your suggestion is how it _should_ be, but this condition is the old condition which has been around forever. IMHO, in the long run `lychee.publicMode` should be nuked completely as this should be equivalent to `lychee.user !== null`.

The point is `lychee.user` is rather new and has only recently introduced by https://github.com/LycheeOrg/Lychee-front/pull/310 while `lychee.publicMode` has been there forever since the dawn of Lychee to distinguish between an unauthenticated user and the authenticated user. This goes back to the time when the only authenticated user was the admin, i.e. when Lychee was single-user.

I don't know which legacy code somewhere somehow sets `lychee.publicMode` for some weird reason to achieve a certain effect. Hence, I would like to keep it (for now) until all code which _sets_ `lychee.publicMode` is gone and we are sure that `lychee.publicMode` is never `true`.

ildyria (approved):
LGTM.

PR #350, opened 2022-11-07
Repairs QR code

nagmat84:
This fixes the broken QR code as noticed by @kamil4.

I assume the comment in the code explains the problem pretty straight forward

```php
// We need this indirection based on a resize observer, because the ready
// callback of the dialog is invoked _before_ the dialog is made visible
// in order to allow the ready callback to make initializations of
// form elements without causing flicker.
// However, for invisible elements `.clientWidth` returns zero, hence
// we cannot paint the QR code onto the canvas before it becomes visible.
```

Of course, an easier fix would have been to simply hard-code the width of the canvas into the JS code. However, this would have required to keep it in sync with style sheet which momentarily makes the canvas to be 300px (width of dialog) - 2 x 12px (padding of dialog) = 276px wide. I decided to implement a little bit more complicated, but hopefully more robust solution.

### DIFF
--- a/scripts/main/album.js
+++ b/scripts/main/album.js
@@ -1256,11 +1256,22 @@ album.qrCode = function () {
 		return;
 	}
 
-	basicModal.show({
-		body: "<div class='qr-code-canvas'></div>",
-		classList: ["qr-code"],
-		readyCB: function (formElements, dialog) {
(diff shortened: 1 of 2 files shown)


ildyria (approved):
Tested. LGTM.

PR #358, opened 2022-11-27
Added smart album - On This Day

aldjordje:
Added smart album - On This Day. Front end.
https://github.com/LycheeOrg/Lychee/issues/1599

### DIFF
--- a/scripts/main/lychee_locale.js
+++ b/scripts/main/lychee_locale.js
@@ -130,6 +130,7 @@ lychee.locale = {
 	STARRED: "Starred",
 	RECENT: "Recent",
 	PUBLIC: "Public",
+	ON_THIS_DAY: "On This Day",
 	NUM_PHOTOS: "Photos",
 
 	CREATE_ALBUM: "Create Album",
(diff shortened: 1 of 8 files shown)


scripts/main/lychee_locale.js
```diff
@@ -130,6 +130,7 @@ lychee.locale = {
 	STARRED: "Starred",
 	RECENT: "Recent",
 	PUBLIC: "Public",
+	ON_THIS_DAY: "On This Day",
```
ildyria, line 133: You will want to modify also the following files:
https://github.com/LycheeOrg/Lychee/tree/master/app/Locale :)

aldjordje: Thanks, I missed that. I'm going to update those files with the help of google translate :)

ildyria: You don't necessarily need to use Google translate, you will notice that there are some untranslated strings still in English left in each languages. :)

aldjordje: You're right. It's safer to leave a native speaker to do a translation from English :)

PR #366, opened 2022-12-19
Fix header if going back from map with all albums

qwerty287:


ildyria (approved):
Untested but looks sane.
LGTM.

PR #367, opened 2022-12-21
Fix bugs

qwerty287:
- Fix sidebar if no license is set, prevents empty `Reuse` category.
- Fix unlocalized strings.
- Fix `null` values shown in EXIF metadata (`editable` was `underfined` but the check only checks `editable === false` and `undefined` results in `true`).
- Fix tags `X` always shown.


ildyria (approved):
LGTM.

PR #369, opened 2022-12-23
Fix various issues on mobile (responsive design)

qwerty287:
- Instead of a contextmenu, use a full screen left sidebar. See screenshot.
- Don't use the `touchend` event. This made all actions unusable for me because it triggers the actions twice: once for `click` and once for `touchend`.

Screenshot 2022-12-23 at 09-40-48 Lychee v4 – Albums


ildyria (approved):
LGTM.
I noticed that contextmenu.config was only used there, so I retired it completely.

PR #378, opened 2023-01-10
1676 - refresh to the album page when album page is opened

cshyam1892:


cshyam1892:
I just realized that the search and setting don't interact either. 
It's just a hack for the album, so we'd need to dig deeper.

cshyam1892:
Looking at this image: 
https://user-images.githubusercontent.com/7428426/196762407-a9a45933-ce95-40ff-b149-b5d29d225610.png

Rather than loading under imageview, shouldn't individual images be under lychee_view_content ?

scripts/main/lychee.js
```diff
@@ -509,7 +509,7 @@ lychee.localizeStaticGuiElements = function () {
 					: lychee.site_copyright_begin + "–" + lychee.site_copyright_end
 		  )
 		: "";
-	footer.querySelector("p.personal_text").textContent = lychee.footer_additional_text;
+	footer.querySelector("p.personal_text").innerHTML = lychee.footer_additional_text;
```
ildyria, line 512: that part is no go as it potentially allows XSS.

cshyam1892: done, I would appreciate some advice on this:

https://github.com/LycheeOrg/Lychee-front/pull/375

cshyam1892:
I've made some changes to the search and settings, and I would be very grateful if you could review my code. I would love to hear your suggestions and feedback. Thank you!

scripts/main/lychee.js
```diff
@@ -1034,14 +1034,23 @@ lychee.load = function (autoplay = true) {
 			$("#sensitive_warning").removeClass("active");
 			if (album.json && albumID === album.json.id) {
 				if (album.isSearchID(albumID)) {
-					// We are probably coming back to the search results from
-					// viewing an image.  Because search results is not a
-					// regular album, it needs to be treated a little
-					// differently.
-					header.setMode("albums");
-					lychee.setMetaData(lychee.locale["SEARCH_RESULTS"]);
+					if ($(".class_settings").length > 0) {
```
ildyria, line 1037: I would rather use
```suggestion
					if ($(".settings_view, .users_view, .sharing_view, .logs_diagnostics_view, .u2f_view").length > 0) {
```
And not use `.class_settings` as it is an empty css selector (i.e. it does not provide style).

cshyam1892: Thank you for taking the time to review. 
I have made the changes.

ildyria (approved):

Issue #380, opened 2023-01-28
WebAuthn.js does not work on older Safari versions


nicokaiser:
This is a copy of [Lychee#1710](https://github.com/LycheeOrg/Lychee/issues/1710)

Since (about) version 4.7, Lychee does not work in older Safari browsers. Safari <= 13 is present on older iOS versions, and most probably some older android devices are also affected.

The culprit seems to be the new WebAuthn.js, which uses private class methods, and which is directly committed into the dist folder (and thus not being transpiled by the Babel task in Lychee-front.

Has this been a deliberate choice, or should the files in dist also be somewhere else and included in the transpilation step?

I‘d happy to help with a PR, but am not yet familiar with the interweaved structure of Lychee and Lychee-front…

nicokaiser:
This definitively affects iOS 12 (which was just recently updated with a security fix, so it is still around on some older devices), but also iOS/iPadOS 15 (which could be updated to iOS 16).

There are some other things that break on "older" browsers, like the "overflow: clip" CSS property and the use of the ResizeObserver, but I'll address these in another PR.

PR #381, opened 2023-01-28
Add WebAuthn.js and add to gulpfile (#380)

nicokaiser:
This PR adds WebAuthn.js from [Lychee](https://github.com/LycheeOrg/Lychee) to Lychee-front and compiles it using Babel, which allows it to be run on older browsers not supporting private class properties.


nicokaiser:
The compiled file should probably added to https://github.com/LycheeOrg/Lychee (when/if this PR gets merged and the submodule is updated)

nicokaiser:
Maybe it would be an idea to compile WebAuthn.js into frontend.js, what do you think?
Currently, [app.blade.php](https://github.com/LycheeOrg/Lychee/blob/master/resources/views/layouts/app.blade.php) includes WebAuthn.js, but this may not be needed...

PR #385, opened 2023-01-28
Hide share button on photos page when share_button_visible is false (#384)

### LINKED ISSUE #384: Share button is shown on photo page even if `share_button_visible` is not set
When `share_button_visible` is disabled in the configuration, the photo view shows the share button, but the button does not show the context menu.
This is confusing for visitors, it should not be visible at all.

nicokaiser:
Fixes #384

The "#button_share" element is being hidden before (`if (!lychee.share_button_visible)` block), however it is shown again in the "Hide buttons if needed" block. It should only be shown if `enable_button_share` and `share_button_visible` are truthy.

### DIFF
--- a/scripts/main/header.js
+++ b/scripts/main/header.js
@@ -534,7 +534,7 @@ header.setMode = function (mode) {
 				const e = $("#button_visibility", "#lychee_toolbar_photo");
 				e.show();
 			}
-			if (!lychee.enable_button_share) {
+			if (!lychee.enable_button_share || !lychee.share_button_visible) {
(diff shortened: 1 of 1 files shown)


ildyria:
> The "#button_share" element is being hidden before (`if (!lychee.share_button_visible)` block), however it is shown again in the "Hide buttons if needed" block. It should only be shown if `enable_button_share` and `share_button_visible` are truthy.

Actually, it should make use of `album.rights.can_share` (or similarly named propert).

ildyria (approved):

Issue #346, opened 2022-10-26
Scroll position should only be restored when navigating from photos back to their album


nicokaiser:
I see that remembering the scroll position is a good thing when navigating from a single photo view back to the album view.
However in other situations it is very distracted and does not make sense, especially if it is permanently saved in localStorage:

1. Open a large album
2. Scroll down
3. Use the "back" link to open the album list
4. Click on the same album again

Alternatively, open the same album via direct link.

I would expect to see the top of the album again, but instead, the bottom of the album is displayed, which is very confusing.

=> Scroll position should only be restored when navigating back from a photo to its album. Otherwise (direct album requests or navigation from parent albums) the position should be set to top.

I don't know if this is doable with the current architecture...

nagmat84:
What you suggest is actually quite easy, because we don't need to remember the scroll position for that at all.

If you open the photo view, the album view is not closed, but remains open in the background. The photo view is implemented as an overlay with an opaque background which covers the album view, but it does not replace the album view. (Indeed, the black background of the photo view is transparent in "TV mode" such that you still see the album view around the big photo in the center.) Hence, we do not need to store the scroll position at all, the browser already does that for us.

The feature to remember and restore the scroll position has actually been added only for the exact purpose which you deem annoying. (Me, too, by the way.) But someone found that particular use case important enough to add this feature.

nicokaiser:
Thanks @nagmat84! Then maybe there should be a configuration option to be able to disable this behaviour. I already added this to Lychee-front but still need to figure out how to add it to the backend (maybe I make it an "advanced" configuration option which needs to be set using the "More" button)...

kamil4:
Things become a bit more difficult with nested albums. True, we don't normally need to record the position of the immediate parent album; however, the position of that album's parent would be lost if one were to push the back button with in the subalbum (the parent album needs to get reloaded then).

In principle though, I agree with you and I think that it shouldn't be too hard to distinguish between clicking on an album vs using the back button. We already do that in the front end when deciding whether to autoplay videos (yes on direct click, no on previous/next).

Personally, I've wondered since this feature was first proposed whether the positions should really be stored as permanently as they are, but I guess I was swayed by the "what if somebody hits the reload" argument...

Another annoying feature of the current implementation is that it doesn't seem to handle changing window geometry correctly. I believe it records an absolute position in pixels, which causes problems e.g. when switching from portrait to landscape on a phone. We should probably record percentages instead...

[...]

PR #391, opened 2023-02-07
Use sessionStorage instead of localStorage for scroll position (#346)

nicokaiser:
With this PR, Lychee saves the scroll position to sessionStorage instead of localStorage, so scroll positions are not kept between tabs/sessions.

The current behaviour can be confusing, especially when opening Lychee after a while, navigating to an album and finding the scroll position at some random value. Using sessionStorage clears the positions when the browser window/tab is closed, while still maintaining the positions while navigating in the same window/tab.

Completely removing the rememberScroll logic is not possible yet, since this would lead to a different, yet still confusing behaviour when navigating between albums and photos, and when using the back/forward buttons of the browser.

So the proper fix for #346 would be to _not_ restore the scroll position when navigating forward by clicking on an item, but restoring the scroll position when either navigating backward or navigating forward using the browser's forward button.

### DIFF
--- a/scripts/main/init.js
+++ b/scripts/main/init.js
@@ -412,7 +412,7 @@ $(document).ready(function () {
 	 */
 	const rememberScrollPage = function () {
 		if ((visible.albums() && !visible.search()) || visible.album()) {
-			let urls = JSON.parse(localStorage.getItem("scroll"));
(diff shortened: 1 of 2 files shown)


ildyria (approved):
LGTM. No strong opinion.
@qwerty287 ?

PR #392, opened 2023-02-08
Fix visibility of share button

qwerty287:
https://github.com/LycheeOrg/Lychee/issues/1752#issuecomment-1422419988


oblomow:
Yes, that fixes the problem. Thanks guys! Quick as always.

d7415 (approved):
Looks good. Untested.

PR #397, opened 2023-04-20
Fix API token dialog box

ildyria:
Fixes https://github.com/LycheeOrg/Lychee/issues/1778


ildyria:
I tested on Debian Firefox, Debian Chrome, Debian Chromium.
On all 3 this worked.

PR #403, opened 2023-06-26
Fixes log link

ildyria:


d7415 (approved):
I haven't tested, but that was the conclusion I came to.

PR #408, opened 2023-09-04
Fix photo count of zero or undefined

evoludolab:
Fixes GUI part of LycheeOrg/Lychee#2002.

If album_decoration==photo, show photo count only if at least one photo in album (same as for album_decoration==all).


ildyria (approved):

ildyria:
This will need a Lychee-front sync later.

